PR #2, opened 2015-04-22
Bugfix: xgboost</lib/native/libhdfs.a: could not read symbols: Bad value

loadingvx:
For Hadoop_CDH distribution (yum installed), there exists no libhdfs.so(dynamic lib) just libhdfs.a(static lib). 
A switch "HADOOP_CDH_BINARY" is added for this situation, it enables static compiling, so xgboost
wil not search for libhdfs.so anymore while running.

### DIFF
--- a/make/dmlc.mk
+++ b/make/dmlc.mk
@@ -22,7 +22,11 @@ endif
 
 ifeq ($(USE_HDFS),1)
 	DMLC_CFLAGS+= -DDMLC_USE_HDFS=1 -I$(HADOOP_HDFS_HOME)/include -I$(JAVA_HOME)/include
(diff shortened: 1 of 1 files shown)


tqchen:
As another comment in the xgboost repo. It would be great if this can be resolved in just dmlc.mk by configuring DMLC_LDFLAGS that xgboost will include.

I think we can do ```-Wl,static``` followed by lhdfs. Also as I recalled correctly, -lhdfs indeed will search for libhdfs.a while libhdfs.so do not exist in search path, can you clarify this?

Thanks

loadingvx:
ok. I will try fix it like this. -Wl,static is really more better than my patch.

PR #3, opened 2015-04-28
Bugfix: libdmlc.a compiling error: fp not define

loadingvx:
Some level typo mistakes in hfs_filesys.cc caused dmlc-core compiling failure.


tqchen:
oops, thanks, forget to check against libhdfs on my dev machine

PR #11, opened 2015-05-22
if worker task failed by abortJob(), the ApplicationMaster process will not be exited.

panhongan:
bugfix


tqchen:
Thanks, I will merge it in for now. However, the original intention of the code was to let the appmaster kill all the containers before it quit. Seems that when a container was killed using stopContainer, onContainerCompleted will not be invoked(thus finishedTasks not updated).

Maybe onContainerStopped will be invoked and can be used to update the state. The current solution also works by killing the app on the client side

PR #22, opened 2015-07-02
add cpplint.py, but “make lint” still depends on pylint

mli:


mli:
also it's a good idea to reuse make lint for other dmlc projects

tqchen:
do not

tqchen:
Once the lint script is ready, we can easily grab it from other repo and reuse it

PR #24, opened 2015-07-15
Concurrency

hotpxl:
simple concurrent queue moved into core.

please check if everything (including style) fits in

### DIFF
--- a/include/dmlc/concurrency.h
+++ b/include/dmlc/concurrency.h
@@ -1,0 +1,49 @@
+/*!
+ * Copyright (c) 2015 by Contributors
+ * \file concurrency.h
+ * \brief thread-safe data structures.
+ * \author Yutian Li
+ */
+#ifndef DMLC_CONCURRENCY_H_
+#define DMLC_CONCURRENCY_H_
+// this code depends on c++11
+#if DMLC_USE_CXX11
+
+namespace dmlc {
+
+/*!
+ * \brief Simple userspace spinlock implementation.
+ */
+class Spinlock {
+ public:
+  Spinlock() = default;
+  /*!
+   * \brief Disable copy and move.
+   */
+  Spinlock(Spinlock const&) = delete;
+  Spinlock(Spinlock&&) = delete;
+  Spinlock& operator=(Spinlock const&) = delete;
+  Spinlock& operator=(Spinlock&&) = delete;
+  ~Spinlock() = default;
+  /*!
+   * \brief Acquire lock.
+   */
+  void lock() noexcept {
+    while (lock_.test_and_set(std::memory_order_acquire)) {
+    }
+  }
+  /*!
+   * \brief Release lock.
+   */
+  void unlock() noexcept {
+    lock_.clear(std::memory_order_release);
+  }
+
+ private:
+  std::atomic_flag lock_ = ATOMIC_FLAG_INIT;
+};
+
+}  // namespace dmlc
+
+#endif  // DMLC_USE_CXX11
+#endif  // DMLC_CONCURRENCY_H_
--- a/include/dmlc/concurrency.h
+++ b/include/dmlc/concurrency.h
@@ -32,16 +32,11 @@ class Spinlock {
   /*!
    * \brief Acquire lock.
    */
-  void lock() noexcept {
-    while (lock_.test_and_set(std::memory_order_acquire)) {
-    }
-  }
+  inline void lock() noexcept;
   /*!
    * \brief Release lock.
    */
-  void unlock() noexcept {
-    lock_.clear(std::memory_order_release);
-  }
+  inline void unlock() noexcept;
 
  private:
   std::atomic_flag lock_ = ATOMIC_FLAG_INIT;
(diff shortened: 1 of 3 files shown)


include/dmlc/concurrency.h
```diff
@@ -0,0 +1,139 @@
...
+      cv_.notify_all();
+    }
+  }
+  /*!
+   * \brief Pop element from the queue.
+   * The element will be copied or moved into the object passed in.
+   * \param e Element popped.
+   * \return Whether the queue is not empty afterwards.
+   */
+  bool Pop(T* rv) {
```
tqchen: style: consider put inline explicitly as in other files. 
Consider move definition to later part of the code, outside the class, see io class. 
So reader can read the declaration with comments clearly

include/dmlc/concurrency.h
```diff
@@ -0,0 +1,139 @@
...
+   */
+  size_t Size() {
+    std::unique_lock<std::mutex> lock{mutex_};
+    return queue_.size();
+  }
+
+ private:
+  std::mutex mutex_;
+  std::condition_variable cv_;
+  std::atomic<bool> exit_now_{false};
```
tqchen: exit now do not have to be atomic, eventual consistency should work here

include/dmlc/concurrency.h
hotpxl: ok.

method definition in header are inline by standard, right?

include/dmlc/concurrency.h
hotpxl: it `exit` is not `atomic`, then it has to be `volatile`. I like shared variables to be more explicit this way

include/dmlc/concurrency.h
tqchen: yes, just to make it explicit.

tqchen:
Thanks for the great work. Here are a few things that are optional, but maybe helpful

* it would be great to explictily add inline to these header functions to indicate that we intend them to be inlined.
* Move the big inline functions out of class body and put them in later part of the header file
   - This helps the user of the library to only read the declaration of the class without read the impl

tqchen:
If @mli or @jermainewang can give it a pass, we can merge it in. thanks!

hotpxl:
Let me add the inline before moving on

mli:
there is a simple disable_copy_and_assign in somewhere, but i think you can define a better one in dmlc/base.h

include/dmlc/concurrency.h
tqchen: now i think exit do not have to be atomic, because memory order is already enforced by the mutex.

hjk41: exit_now_ need to be atomic or volatile (which is a weaker guarantee) in
order to prevent compiler optimization that caches exit_now_ in register,
preventing other threads to know the status of exit_now_. Of course we
probably don't need to worry about that with -O2, but it is still the right
thing to do.

On Wed, Jul 22, 2015 at 11:14 AM, Tianqi Chen <notifications@github.com>
wrote:

> In include/dmlc/concurrency.h
> <https://github.com/dmlc/dmlc-core/pull/24#discussion_r35178097>:
>
> > +    exit_now_.store(true);
> > +    cv_.notify_all();
> > +  }
> > +  /*!
> > +   * \brief Get the size of the queue.
> > +   * \return The size of the queue.
> > +   */
> > +  size_t Size() {
> > +    std::unique_lock<std::mutex> lock{mutex_};
> > +    return queue_.size();
> > +  }
> > +
> > + private:
> > +  std::mutex mutex_;
> > +  std::condition_variable cv_;
> > +  std::atomic<bool> exit_now_{false};
>
> now i think exit do not have to be atomic, because memory order is already
> enforced by the mutex.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/dmlc/dmlc-core/pull/24/files#r35178097>.
>



-- 
HONG Chuntao
System Research Group
Microsoft Research Asia

tqchen: I know that if mutex is not presented, then exit_now_ need to be atomic to ensure the memory aquire release order. But in such case memory aquire release order is already satisfied due to mutex. what @hjk41  suggested is that volatile is needed. Which I do agree. Though I do not know if compiler will optimize class member(it is reasonable to do so for local variables)

mli: adding std::atomic<bool> is reasonable since we often use -O3. i remember in an old gcc version -O3 includes several dangerous optimization such as remapping registers. it once took me several days for debugging.

hjk41: Right. There is no guarantee that compiler will *not* do this kind of
aggressive optimization. The standard does not forbid this.

On Wed, Jul 22, 2015 at 11:39 AM, Mu Li <notifications@github.com> wrote:

> In include/dmlc/concurrency.h
> <https://github.com/dmlc/dmlc-core/pull/24#discussion_r35179004>:
>
> > +    exit_now_.store(true);
> > +    cv_.notify_all();
> > +  }
> > +  /*!
> > +   * \brief Get the size of the queue.
> > +   * \return The size of the queue.
> > +   */
> > +  size_t Size() {
> > +    std::unique_lock<std::mutex> lock{mutex_};
> > +    return queue_.size();
> > +  }
> > +
> > + private:
> > +  std::mutex mutex_;
> > +  std::condition_variable cv_;
> > +  std::atomic<bool> exit_now_{false};
>
> adding std::atomic is reasonable since we often use -O3. i remember in an
> old gcc version -O3 includes several dangerous optimization such as
> remapping registers. it once took me several days for debugging.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/dmlc/dmlc-core/pull/24/files#r35179004>.
>



-- 
HONG Chuntao
System Research Group
Microsoft Research Asia

tqchen: OK, i guess volatile gives weaker gaurantee and still works for this case

PR #26, opened 2015-07-17
Added DMLC_DECLARE_TRAITS(is_pod, int8_t, true);

rodrigob:
Fixes compilation on g++-4.8 and clang++-3.5.

### DIFF
--- a/include/dmlc/type_traits.h
+++ b/include/dmlc/type_traits.h
@@ -47,6 +47,7 @@ struct has_saveload {
 //! \cond Doxygen_Suppress
 // declare special traits when C++11 is not available
 #if DMLC_USE_CXX11 == 0
+DMLC_DECLARE_TRAITS(is_pod, char, true);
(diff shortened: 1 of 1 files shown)


tqchen:
Please also remove the line for int8_t since I guess char will be more often used, and seems MSVC indeed use char for int8_t

rodrigob:
This is exactly what I was worried about. Simply adding "char" is a not a cross-platform solution, as would be simply removing int8_t. We should make this compiler dependent, or have a good way to detect (at compilation time) of char == int8_t.

tqchen:
Alright, I think it is fine anyway, since traits is only needed for non c++11 compile mode, and MSVC is by default c++11 and this section will not be enabled. I will merge it in for now. 

Thanks!

tqchen:
We can remove int8_t for now since it is not usually used in most cases. In common case(c++11 enabled) the library will always compile correctly

PR #36, opened 2015-08-18
update strtonum

krishnap25:
Updated strtonum to work with unsigned long long

### DIFF
--- a/src/data/strtonum.h
+++ b/src/data/strtonum.h
@@ -178,6 +178,13 @@ class Str2T<uint64_t> {
   }
 };
 
+template<>
+class Str2T<unsigned long long> {
+ public:
+  static inline unsigned long long get(const char * begin, const char * end) {
+    return strtoint<unsigned long long>(begin, NULL, 10);
+  }
+};
 
 template<>
 class Str2T<float> {


tqchen:
@mli  can you look into this? There is an interesting issue about unsigned long long, that makes some of things break in certain platform. It should have been covered by uint64_t. I am not sure why this specialization is needed  here.

mli:
hi Krishna, 

this is due to gcc4.8's bug. please make a pull for ps-lite, or follow this commit:

https://github.com/dmlc/ps-lite/commit/a47fe951c55a089efadb6ae8df866aa3fe031f8a

adding Str2T<unsigned long long> will leads to compilation failure on recent gcc and clang.

krishnap25:
I'm getting the same bug even though I use gcc 4.9.1 and 4.9.2, and use the
latest commit of ps-lite. What do I do now?

One alternative is to guard Str2T using the __GNUC__ macro.

-Krishna Pillutla

mli:
the bug is due to some compilers think `unsigned long long` and `uint64_t` are different types, but the others think they are the same one. dmlc-core uses uint64_t, so you may change your codes to follow this conversion.

PR #38, opened 2015-08-23
[concurrency-fix-for-gcc-4.8] member initialization doesn't work on G++ 4.8

hotpxl:
I don't know why but the old syntax doesn't work on my G++ 4.8. I had to revert to the old way.

Please review and see if it breaks anything.


tqchen:
lgtm, thanks.

PR #43, opened 2015-09-17
Parameter serialization

Iroul:
The const cast is ugly tho. However because void* head is used everywhere I guess we will live with it.


tqchen:
Was merged  in #44

PR #57, opened 2015-10-06
add cmake 

yajiedesign:
add cmake file ,Compile with vs2013

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,34 @@
+﻿cmake_minimum_required(VERSION 2.8.7)
+
+project(dmlccore C CXX)
+
+list(APPEND CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake/Modules)
+
+include(cmake/Utils.cmake)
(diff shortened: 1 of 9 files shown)


tqchen:
Thanks for the contribution! Please fix the compile and lint errors of travis here https://travis-ci.org/dmlc/dmlc-core/builds/84019619

yajiedesign:
sorry more commit, my epylint not work.

tqchen:
Thanks for the contribution! Can you check what is the difference between lint and lint.py Is it possible that we combine the two and do them in one file?

yajiedesign:
@tqchen you say lint.py and lint3.py?lint.py work in py2 and lint3.py work in py3.

tqchen:
I guess we can make a py2, py3 compatible version of lint, and simply import lint in lint3.py ?

tqchen:
@yajiedesign Just a reminder on this. I think we only need to keep the content in lint3.py rename it to lint.py and use it for both lint in python3 and python3. The code should be compatible in both versions

tqchen:
It would be great if you can submit a PR to update this

PR #75, opened 2015-11-26
[PARSER] allow registry pattern  to add new parsers standalone from DMLC

tqchen:
@mli @hjk41

### DIFF
--- a/include/dmlc/registry.h
+++ b/include/dmlc/registry.h
@@ -188,8 +188,8 @@ class FunctionRegEntryBase {
  */
 #define DMLC_REGISTRY_ENABLE(EntryType)                                 \
   template<>                                                            \
-  Registry<EntryType> *Registry<EntryType>::Get() {                     \
-    static Registry<EntryType> inst;                                    \
+  Registry<EntryType > *Registry<EntryType >::Get() {                   \
+    static Registry<EntryType > inst;                                   \
     return &inst;                                                       \
   }                                                                     \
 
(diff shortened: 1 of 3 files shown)


tqchen:
@mli can you review and merge this?

include/dmlc/registry.h
```diff
@@ -188,8 +188,8 @@ class FunctionRegEntryBase {
  */
 #define DMLC_REGISTRY_ENABLE(EntryType)                                 \
   template<>                                                            \
-  Registry<EntryType> *Registry<EntryType>::Get() {                     \
-    static Registry<EntryType> inst;                                    \
+  Registry<EntryType > *Registry<EntryType >::Get() {                   \
```
hjk41: why do we have blackspace after EntryType? is it a coding convention?

hjk41:
looks good to me

include/dmlc/registry.h
tqchen: EntryType could contain something that ends with ```>```, which I am not sure recognized by old c++ compiler(should be ok for c++11 compatible ones) 

```vector<vector<int>> ``` vs ```vector<vector<int> > ```

tqchen:
Ideally this allow plug in of user customized parser into dmlc modules without actually putting code in dmlc codebase

hjk41:
I see. Looks great.

On Tue, Dec 1, 2015 at 1:32 PM, Tianqi Chen <notifications@github.com>
wrote:

> Ideally this allow plug in of user customized parser into dmlc modules
> without actually putting code in dmlc codebase
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/dmlc/dmlc-core/pull/75#issuecomment-160858191>.
>



-- 
HONG Chuntao
System Research Group
Microsoft Research Asia

PR #77, opened 2015-12-09
Some minor patches to improve S3 reliability:

ylow:
- CURL_NOSIGNAL should be set when used in a multithreaded setting
 - Get better errors when curl_multi is used
 - Add an explicit Close() method call to the writestream
 - Add an explicit SetCredential method
 - Add support for creation of 0 byte files

### DIFF
--- a/src/io/s3_filesys.cc
+++ b/src/io/s3_filesys.cc
@@ -70,7 +70,7 @@ struct XMLIter {
  * \param len the length of data
  * \return the encoded string
  */
-std::string Base64(unsigned char md[], unsigned len) {
+static std::string Base64(unsigned char md[], unsigned len) {
   // encode base64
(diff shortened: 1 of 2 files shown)


mli:
hi yucheng,

many thanks for the PR. the s3 unreliability bug bothered me for a while. for example, using 10 machines to read a single file, the probability that 1 machine is blocked within 1h is pretty high. can your PR solve this issue?

ylow:
I do find the CURLOPT_NOSIGNAL does help with stability issues when used in a multi-threaded setting. But it is hard to say whether this will fix that s3 issues without actually trying it out.

mli:
thanks, i'll run a large ec2 job to test the stability later. 

LGTM now, @tqchen please also have a look

PR #81, opened 2015-12-29
support hdfs accessing across different clusters

sharkdtu:
At present, the dmlc-core only support default HDFS cluster accessing. However, we often need to access multiple HDFS clusters for data and model respectively in many different scenarios. So this patch aims to support hdfs accessing across different clusters. We can config the data or model througth absolute path,  such as:

data=hdfs://host:port/path/to/data
model_out=hdfs://otherhost:otherport/path/to/model

### DIFF
--- a/src/io.cc
+++ b/src/io.cc
@@ -27,18 +27,18 @@
 
 namespace dmlc {
 namespace io {
-FileSystem *FileSystem::GetInstance(const std::string &protocol) {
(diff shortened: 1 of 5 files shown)


tqchen:
Thanks for the contribution, please fix the lint error here https://travis-ci.org/dmlc/dmlc-core/builds/99229593

You can reproduce locally via ```make lint```

PR #94, opened 2016-01-15
fix

tqchen:


kr-deepak:
Hi Chen,

Trying to build xgb using mingw on windows 7. Some problems that I am facing:

First of all when I download the complete zip, dmlc-core and rabit-core dont get downloaded. I had to download them separately becuase cp make/mingw64.mk config.mk was saying no makefile.

Then when I run cp make/mingw64.mk config.mk; make the it exits with  [make ** error build/learner.o Error 1]. 
If I run cp make/mingw64.mk config.mk; make -j4 it just hangs up for more than 30 minutes. Is it supposed to take too long to compile. There is no error here (why so?)

I was doing as you suggested in http://xgboost.readthedocs.org/en/latest/build.html.

Please let me know what I need to do to get the make running.

tqchen:
please open a issue here https://github.com/dmlc/xgboost/issues

PR #98, opened 2016-01-20
fix yarn for windows

hjk41:
1. add maven project file
2. fix some problems on windows
TODO: classpath for workers still has some problem


hjk41:
@tqchen

PR #100, opened 2016-01-25
Fix the segmentation fault with ostream caused by invalid pointer access.

hang-qi:
This pull request fixes a segmentation fault in current `ostream`.

In the current code base, `iostream` is used as follows:

    Stream *fs = Stream::Create("hdfs:///test.txt", "w");
    dmlc::ostream os(fs);       // <-- os does not own the fs pointer.
    os << "hello world" << std::endl;
    delete fs;                  // <-- the fs pointer gets deleted.
    // SIGSEGV when os is destructed.

When destructing the `os` object, it tries to sync the buffer which accesses the memory that has been deleted. When I ran `test/test_iostream` with HDFS, it caused a segmentation fault!

To fix this, in this pull request, I changed the interface and make `ostream` and `istream` instances own the underlying Streams. Specifically, they create the stream in the constructor and release the stream (and therefore close the file) at the destruction. The usage becomes:

    dmlc::ostream os("hdfs:///test.txt");
    os << "hello world" << std::end;
    // file will be closed when os is destructed.

The inline docxygen docs have been updated accordingly.

Discussion: later it may be better to add a `close()` function to allow closing the file explicitly. And it seems to me that it would make more sense if `In/OutBuf`inherits from `filebuf` instead of `streambuf`. However, this will involve much bigger changes. So, to avoid messing up the base with a huge PR, I am submitting this PR at the current point. Please let me know what you think.

### DIFF
--- a/include/dmlc/io.h
+++ b/include/dmlc/io.h
@@ -210,24 +210,26 @@ class InputSplit {
 };
 
 /*!
- * \brief a std::ostream class that can can wrap Stream objects,
+ * \brief a std::ostream class that wraps a Stream object,
  *  can use ostream with that output to underlying Stream
  *
  * Usage example:
  * \code
  *
- *   Stream *fs = Stream::Create("hdfs:///test.txt", "w");
- *   dmlc::ostream os(fs);
+ *   dmlc::ostream os("hdfs:///test.txt");
  *   os << "hello world" << std::endl;
- *   delete fs;
  * \endcode
  */
 class ostream : public std::basic_ostream<char> {
  public:
   /*!
    * \brief construct std::ostream type
-   * \param stream the Stream output to be used
+   * \param uri the uri of the input currently we support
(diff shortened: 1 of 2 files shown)


tqchen:
Yes, I am aware of this. The main reason was because ostream need to be destructed before the stream get  deleted. One way to enforce that is via
```
std::unique_ptr<Stream> fs(Stream::Create("hdfs:///test.txt", "w");
dmlc::ostream os(fs);       // <-- os does not own the fs pointer.
os << "hello world" << std::endl;
// force reset the file pointer so things are flushed here
os.reset(nullptr);

```

tqchen:
I understand this is a bit tricky as originally istream and ostream are not intended to take ownership of Stream, which makes it a bit bad API design as it is context dependent.

I will close this PR for now, mainly because of some of the backward compatiblity issue, we should definitely update the doc example to ask user to explicitly do so. This problem do not present in case of istream

PR #104, opened 2016-01-27
allow specifying return type for FunctionReg

hjk41:
Some functions like SliceChannel produces multiple outputs. This fix will enable those functions to specify the output type, so it can be Symbol or Symbol[]. I will create a PR for that on mxnet.
@tqchen @mli


tqchen:
sounds good, we can merge it in after you do the squash

PR #110, opened 2016-02-17
[YARN] Specify node's `host` in container env.

BaiGang:
The current implementation gets ip/host via querying network adaptor info. This will raise issues when the ip/host got is not accessible from outside the container. To fix this, we can use the YARN node host of the container, which is guaranteed to be accessible from outside the container, i.e. other containers and the YARN client node.

This commit just set the node host to the container's env. To use the node host, we can get the env from the application program, such as those based on ps-lite, executed on YARN.


BaiGang:
Modified titile since this PR is related to yarn.

PR #112, opened 2016-02-20
add fault tolerance

seusofthd:
1. add fault tolerance for ApplicationMaster.java. If the container get errors when running the code, add it to blacklist.
2. add ctrl+c signal response for Client.java

### DIFF
--- a/yarn/src/main/java/org/apache/hadoop/yarn/dmlc/ApplicationMaster.java
+++ b/yarn/src/main/java/org/apache/hadoop/yarn/dmlc/ApplicationMaster.java
@@ -103,6 +103,9 @@ public class ApplicationMaster {
     private final Collection<TaskRecord> killedTasks = new java.util.LinkedList<TaskRecord>();
     // worker environment
     private final Map<String, String> env = new java.util.HashMap<String, String>();
+	
+	//add the blacklist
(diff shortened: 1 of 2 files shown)


yarn/src/main/java/org/apache/hadoop/yarn/dmlc/ApplicationMaster.java
```diff
@@ -300,6 +303,22 @@ private synchronized void submitTasks(Collection<TaskRecord> tasks) {
         }
     }
 
+
+
+	private synchronized void launchDummyTask(Container container){
```
tqchen: Alignment, please use space instead of tab

yarn/src/main/java/org/apache/hadoop/yarn/dmlc/Client.java
```diff
@@ -253,7 +253,7 @@ private void run(String[] args) throws Exception {
         }
         this.initArgs(args);
         // Create yarnClient
-        YarnClient yarnClient = YarnClient.createYarnClient();
+		final YarnClient yarnClient = YarnClient.createYarnClient();
```
tqchen: tab?

seusofthd:
Please review the changes according to your comments.

Issue #111, opened 2016-02-19
[Discussion] Unified Tracker


tqchen:
This is the proposal for a unified tracker script that is used to launch all DMLC jobs. I will post all the necessary parameters here, and please respond the discussion if we want to iterate. Previously the tracker script of DMLC job submission has been separated into different python files. While this can be convenient, it also makes user hard to find a central command instruction  to submit all kinds of jobs. So I am proposing a new centralized script `dmlc-submit` that will submit the jobs.

This discussion thread will eventually becomes the document as well as spec for the submission script. I am hoping we can do it in one to two weeks

Let us also discuss parameters, I think we could avoid using abbreviations and enforce full names, to make submission script more clear, as in other job submission scripts like 

Some of these parameters could be set from environment variables, with prefix DMLC.
## Parameters
- `--num-workers` integer, required
  - Number of workers in the job.
- `--num-servers`` integer, default 0. Number of servers in the job.
- `--worker-cores` integer, default 1 . Number of cores needed to be allocated for worker job.
- `--server-cores` integer, default 1, Number of cores needed to be allocated for server job.
- `--worker-memory` string, integer + [g|m], default 1g, Memory needed for worker job.
- `--server-memory` string, integer + [g|m], default 1g, Memory needed for worker job.
- `--mode` string, 'mpi' or 'yarn' or 'local', default to ${DMLC_SUBMIT_MODE}, submission mode.

tqchen:
@mli @hjk41

hjk41:
https://bitbucket.org/hjk41/popcorn
This is the Yarn code we are using to start different jobs on our clusters,
including MPI jobs and some other parameter server applications. We are
still working actively to improve it. If you are interested, I think we can
make it part of DMLC.

On Sun, Feb 21, 2016 at 12:54 AM, Tianqi Chen <notifications@github.com>
wrote:

> @mli <https://github.com/mli> @hjk41 <https://github.com/hjk41>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/dmlc/dmlc-core/issues/111#issuecomment-186650758>.
>



-- 
HONG Chuntao
System Research Group
Microsoft Research Asia

tqchen:
@hjk41  It seems to be interesting. I think the goal is to make one unified tracker that handles most job  submission tasks. And make one application master that satisfies all the jobs. 

[...]

PR #113, opened 2016-02-24
[TRACKER] refactor tracker

tqchen:


tqchen:
c.f. #111

tqchen:
@mli  can you please review this, let us aim to merge this soon

tqchen:
Summary of changes:
- Moved most python script to tracker/dmlc_tracker
- Moved yarn to tracker/yarn
- Moved yarn/run_hdfs_prog.py to tracker/dmlc_tracker/launcher.py

Added dmlc-submit for submitting all jobs. Removed abbreviations in arguments and only support full names(this might affect previous instructions, and need to be updated)

tracker/dmlc_tracker/mpi.py
```diff
@@ -0,0 +1,44 @@
...
+        else:
+            cmd = 'mpirun -n %d --hostfile %s ' % (nworker + nserver, args.mpi_host_file)
+
+        pass_envs['DMLC_MODE'] = args.mode
+        for k, v in pass_envs.items():
+            # for mpich2
+            if args.mode == 'mpich2':
+                cmd += ' -env %s %s' % (k, v)
+            else:
+                cmd += ' -x %s=%s ' % (k, v)
```
mli: scripts to check mpich or openmpi

```python
    outfile="/tmp/mpiver"
    os.system("mpirun 1>/tmp/mpiver 2>/tmp/mpiver")
    with open (outfile, "r") as infile:
        mpi_ver = infile.read()
    cmd = ''
    if 'Open MPI' in mpi_ver:
        for k, v in envs.items():
            cmd += ' -x %s=%s' % (k, str(v))
    elif 'mpich' in mpi_ver:
        for k, v in envs.items():
            cmd += ' -env %s %s' % (k, str(v))
    else:
        raise Exception('unknow mpi version %s' % (mpi_ver))
```

tqchen: Currently, what we do is simply ask user to specify mpich or openmpi

mli: but not every user know there are several mpi versions, since all provide `mpirun`

tracker/dmlc_tracker/tracker.py
```diff
@@ -367,7 +374,7 @@ def slave_envs(self):
             return {'DMLC_PS_ROOT_URI': self.hostIP,
                     'DMLC_PS_ROOT_PORT': self.port}
 
-def submit(nworker, nserver, fun_submit, hostIP = 'auto', pscmd = None):
+def submit(nworker, nserver, fun_submit, hostIP='auto', pscmd=None):
     if hostIP == 'auto':
         hostIP = 'ip'
     if hostIP == 'dns':
```
mli: please consider this scripts which avoids to pick the 127* ip

```bash
    elif hostIP == 'ip':
        from socket import gaierror
        try:
            hostIP = socket.gethostbyname(socket.getfqdn())
        except gaierror:
            logging.warn('gethostbyname(socket.getfqdn()) failed... trying on hostname()')
            hostIP = socket.gethostbyname(socket.gethostname())
        if hostIP.startswith("127."):
            s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
            # doesn't have to be reachable
            s.connect(('10.255.255.255', 0))
            hostIP = s.getsockname()[0]
```

tracker/dmlc_tracker/mpi.py
tqchen: I will adapt this code

mli:
can you also consider to add the ssh launcher? https://github.com/dmlc/ps-lite/blob/master/tracker/dmlc_ssh.py. only `ssh` is required.

the only difference to others is that, when `--sync-dir dir` is provided, it will rsync the current directory into the remote machine's dir

tqchen:
@mli  I am considering for you to add ssh  options after this PR get merged in

mli:
that also works. the only reason is that i'm not sure if `--sync-dir` is a proper name. or we can automatically do that by 

1. get dir=`pwd` from the scheduler,
2. if dir exists on the worker machine, run rsync
3. if not, rsync dir to remote machine's /tmp/rand-dir

tqchen:
I think we want to avoid doing sync directly, --rsync-dir is a good name

tqchen:
Pushed another commit with fixed your other comments https://github.com/tqchen/dmlc-core/commit/4844d2f4754f5336fdd5ea04eacd89bf3db4ead8

@mli

tqchen:
Pushed a commit to rename --mode to --cluster as discussed with @mli

tqchen:
merge this in for now, still waiting verification from @seusofthd

seusofthd:
@tqchen the local mode works fine but there are some problems with yarn mode. I need to set the -env parameter, how to set it? 
My command is here:
./dmlc-submit --cluster yarn --num-workers 1 --num-servers 1 --host-ip 192.168.1.14 --yarn-app-dir /home/dihuang/mxnet/mxnet/dmlc-core/yarn/ --hdfs-tempdir /tmp/ python noisy_py.train

The error info is here:
 get_opts
    'environment variable DMLC_SUBMIT_CLUSTER')
RuntimeError: --cluster is not specified, you can also specify the default behavior via environment variable DMLC_SUBMIT_CLUSTER

tqchen:
This is quite strange, as you already specified the cluster.
You can add envs by
```
--env name=value
```

tqchen:
BTW, note that the YARN's appMaster need also change to the newest version, to be able to accept the allocations for both server and worker resources separated, your previous PR has been merged into this merge

tqchen:
can you look a bit into it on what is going on? thanks!

seusofthd:
But there are a lot of environment variables to set. I also note that the run_hdfs_prog.py is removed, does it mean that I have to set all of those hdfs related variables manually in --env?

tqchen:
it is changed to launcher.py which is still used

seusofthd:
@tqchen yarn test passed. We have to specify DMLC_JOB_CLUSTER, DMLC_CPU_VCORES, DMLC_MEMORY_MB in the --env parameter. It seems that setting the --worker-cores --worker-memory --server-cores --server-memory does not work. Since we have specified the --cluster yarn in the previous paramter, I think you can export yarn to the system env variables in launch.py, saving the trouble to specify DMLC_JOB_CLUSTER in --env

tqchen:
These parameters are updated to DMLC-Worker-cores etc. you will need to use
the latest yarn from the repo instead of your version. Or update the app
Master on your side

seusofthd:
OK

tqchen:
sorry about the breaking changes, this is mainly for separate resource specification between work process and server process

PR #115, opened 2016-02-29
Dihuang@tusimple

seusofthd:
@tqchen Fix the bugs for Ctrl+C signal as you mentioned. Using SignalHandler will report several warnings at the compile time which is caused by the package itself. Just ignore it.


tqchen:
please do a squash and I will merge it in. Thanks!

tracker/yarn/src/main/java/org/apache/hadoop/yarn/dmlc/ApplicationMaster.java
```diff
@@ -601,7 +601,7 @@ private synchronized void onContainersCompleted(List<ContainerStatus> status) {
                         return;
                     }
                 } catch (Exception e) {
-                    LOG.warn(e.getMessage());
+                        LOG.warn(e.getMessage());
```
tqchen: seems to be extra space or tab?

tracker/yarn/src/main/java/org/apache/hadoop/yarn/dmlc/Client.java
```diff
@@ -34,6 +34,9 @@
 import org.apache.hadoop.yarn.util.ConverterUtils;
 import org.apache.hadoop.yarn.util.Records;
 
+import sun.misc.Signal;
```
tqchen: does this mean the signal handler is sun JVM dependent?

seusofthd:
yes. On my machine I just use openjdk. It can compile.

tqchen:
see https://xgboost.readthedocs.org/en/latest/dev-guide/contribute.html#how-to-combine-multiple-commits-into-one

tqchen:
@seusofthd can you merge your commits? Thanks

seusofthd:
Oh... Ok, I forgot to merge. Let me merge it when I come to office tmr

seusofthd:
@tqchen I followed the instructions on the tutorial and solved all the conflicts, but it seems that it did not work...

tqchen:
could due to there is a merge back operation in your log. One thing you can do is to backup your code, reset to a old version and copy it back. but this can be quite dangerous so be careful to backup first

seusofthd:
ok...

tqchen:
i.e. start from a clean fork Thanks for all the effort

seusofthd:
This would be a easier way i think. I will start from a clean fork then.

tqchen:
any update on this?

seusofthd:
@tqchen the system now runs smoothly in our cluster. I haven't find any other problems yet. Maybe we can write the label scheduling response in appmaster when cdh added label scheduling in their latest version

tqchen:
I mean, did you ever PRed this from another fork?

seusofthd:
@tqchen I remember I've already done a clean fork before and you've merged it

PR #120, opened 2016-03-12
std::chrono in timer.h

WojciechMigda:
Retrieve current timestamp with std::chrono API.

### DIFF
--- a/include/dmlc/timer.h
+++ b/include/dmlc/timer.h
@@ -7,6 +7,12 @@
 #ifndef DMLC_TIMER_H_
 #define DMLC_TIMER_H_
 
+#include "base.h"
+
+#if DMLC_USE_CXX11
+#include <chrono>
+#endif
+
 #include <time.h>
 #ifdef __MACH__
 #include <mach/clock.h>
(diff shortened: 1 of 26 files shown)


WojciechMigda:
sorry, it was meant for my fork

tqchen:
This was actually a good PR, I think I am going to merge this in

tqchen:
It would be great if you can help squash and fix the lint error

WojciechMigda:
Oh, but please double check if this is what you'd expect to have here. I didn't spend significant amount of time trying to understand the context where these timings are used for and this is simply stub I used for topcoder. I have neither analysed the hierarchy of #defines and correctness on win platforms. Do you have that covered in your testing environment?
I'll have a look into the lint error in the meantime.

tqchen:
I see, it is only used to get relative timing, so the chrono is fine. would be nice if you can submit a separate PR just for the timer, but no hurries

WojciechMigda:
Lint is complaining about the header order? If so, how do you enforce that `DMLC_USE_CXX11` is `#defined` before conditional `<chrono>` inclusion occurs if local headers cannot appear before the system ones?

tqchen:
I think rabit's Timer is less frequently used and will be eventually switch to dmlc's timer so we are fine there

PR #135, opened 2016-06-01
Fix the openmp flag bug

jennyzhang0215:
Change NO_OPENMP flag in dmlc.mk to USE_OPENMP
Solve the issues from https://github.com/dmlc/mxnet/issues/1579, https://github.com/dmlc/mxnet/issues/2078

### DIFF
--- a/make/dmlc.mk
+++ b/make/dmlc.mk
@@ -10,7 +10,7 @@ ifndef LIBJVM
 	LIBJVM=$(JAVA_HOME)/jre/lib/amd64/server
 endif
 
-ifndef NO_OPENMP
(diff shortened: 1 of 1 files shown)


sxjscience:
@tqchen, @pluskid Would this fix break compatibility of other packages? I find some OS X users of MXNet have reported building erros related to OpenMP.

pluskid:
It should be fine, as `USE_OPENMP` is set to 0 by default in `osx.mk`. 

(Although it is still quite mysterious why the original way cause problems.)

sxjscience:
The compilation problem should be solved by https://github.com/dmlc/mxnet/pull/2407

PR #161, opened 2016-07-27
In line 149, change {98, 48} to [98, 48].

yxzf:
Python 2.6 has no support to this set literal  {}, and I have changed {98, 48} to [98, 48] @Earthson


Earthson:
@tqchen

PR #173, opened 2016-09-18
[LUA] Add Lua module to do seamelss integration of lua and torch

tqchen:
This is an optional header to interface easier with Lua, allows easier embedding of Lua programs in DMLC projects when torch and lua is available

@piiswrong  @mavenlin 

- Support automatic type conversion back and forth, call lua function directly with mixed C++ and LuaRef types
- Convert return value of functions back
- Access lua table from c++ side easily
- More reliable thread checking

Restrictions
- Only the return first value of the lua function is returned

Example:
```c++
#include <dmlc/lua.h>

void test_torch() {
  dmlc::LuaState* lua = dmlc::LuaState::ThreadLocalState();
  lua->Eval("require 'torch'");
  std::vector<float> data = {1, 2, 3};
  auto storage = dmlc::THStorageSharedData(lua, &data[0], data.size());
  auto add1 = lua->Eval(
      "return "
      "function (s)"
      "  t = torch.FloatTensor(s)"
      "  t:add(1) "
      "end");
  // do add 1 on the C++ dataset
  add1(storage);
  auto print = lua->Eval("return function(x) print(unpack(x)) end");
  print(data);
}

void test_lua() {
  dmlc::LuaState* lua = dmlc::LuaState::ThreadLocalState();
  lua->Eval("require 'torch'");
  lua->Eval("require 'nn'");
  // convert to {1,2,3}
  auto tbl = lua->Convert(std::vector<int>{1,2,3});
  // convert to lua table
  auto tbl2 = lua->Convert(std::unordered_map<std::string, int>{
      {"data", 1},
      {"value", 11}
    });
  auto x = lua->Eval("return 1");
  auto f = lua->Eval("return function(x) return x*x end");
  tbl.SetField("myfunction", f);
  // fetch value from torch table
  auto f2 = tbl["myfunction"];
  // lua function can be directly called from c++
  LOG(INFO) << f2(std::move(tbl2["value"]));
}

int main() {
  test_lua();
  test_torch();
  return 0;
}
```

### DIFF
--- a/include/dmlc/lua.h
+++ b/include/dmlc/lua.h
@@ -1,0 +1,800 @@
+/*!
+ *  Copyright (c) 2016 by Contributors
+ * \file lua.h
+ * \brief C++11 header only interface to easily interact with Lua and Torch.
+ *  This code is evolved from torch plugin code for MXNet.
+ *
+ *  This header will require Torch and Lua to be presented, do not include.
+ *
+ * \author Junyuan Xie, Min Lin, Tianqi Chen
+ *
+ * \code
+ *
+ * // Example code to use the lua module.
+ * dmlc::LuaState* lua = dmlc::LuaState::ThreadLocalState();
+ * // vectors converts automatically to lua table.
+ * auto tbl = lua->Convert(std::vector<int>{1,2,3});
+ * // use eval to get lua reference, this is a function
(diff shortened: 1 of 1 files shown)


tqchen:
@piiswrong if you are ok, let us merge this in

PR #180, opened 2016-10-09
chunked input split for global shuffling

gengyifeng:
With this PR and small changes in data iterators, we can achieve global shuffling. Current implementation does shuffling within an chunk, but a record can not be randomly placed across chunks. The idea for global shuffling is to create more chunks and shuffle chunks first, then shuffle records within a chunk. So the records are shuffled globally.


tqchen:
Thanks! This is merged

PR #185, opened 2016-11-08
Support debug build.

EricFisher:
The patch enables dmlc-core a debug build  same as mxnet.


CMakeLists.txt
```diff
@@ -67,7 +67,12 @@ else(MSVC)
   check_cxx_compiler_flag("-std=c++0x"    SUPPORT_CXX0X)
   check_cxx_compiler_flag("-msse2"        SUPPORT_MSSE2)
   check_cxx_compiler_flag("-openmp"       SUPPORT_OPENMP)
-  set(CMAKE_C_FLAGS "-O3 -Wall -msse2 -Wno-unknown-pragmas -std=c++0x -fPIC")
+  set(CMAKE_C_FLAGS "-Wall -msse2 -Wno-unknown-pragmas -std=c++0x -fPIC")
+  if(NDEBUG)
```
cjolivier01: NDEBUG is set in C/C++ compile, but I am not sure it's set with cmake.  Normally, the check is if CMAKE_BUILD_TYPE is "Debug"
-O3 should be automatic:
http://stackoverflow.com/questions/34302265/does-cmake-build-type-release-imply-dndebug

PR #190, opened 2016-11-21
change parsing behavior to more strict checking

piiswrong:


tqchen:
please fix the travis script, the same pip install --user issue

PR #192, opened 2016-11-27
print value and stack trace

piiswrong:


piiswrong:
@tqchen

include/dmlc/logging.h
```diff
@@ -14,6 +14,10 @@
 #include <stdexcept>
 #include "./base.h"
 
+#if DMLC_LOG_STACK_TRACE
+#include <execinfo.h>
```
khotilov: MinGW (at least the versions 4.9.3 and 6.2.0 that I've got on my machine) doesn't seem to have execinfo.h

include/dmlc/base.h
```diff
@@ -37,6 +37,14 @@
 #define DMLC_LOG_CUSTOMIZE 0
 #endif
 
+/*!
+ * \brief Wheter to print stack trace for fatal error,
+ * enabled on linux when using gcc.
+ */
+#if (!defined(DMLC_LOG_STACK_TRACE) && defined(__GNUC__))
```
khotilov: A check for `&& !defined(__MINGW32__)` or `&& !defined(WIN32)` might be needed. I assume that 64 bit versions have those defined too.

tqchen:
Sorry I did not raise this up when it get merged. But it seems to be better to do everything via macro here to enforce code inlining and avoid the layer of additional inline function call. Consider instead

```c++
#define CHECK_NE(x, y)  \
     CHECK(x!=y) << x << " vs "  << y
```

tqchen:
@piiswrong

piiswrong:
You can't print the value of x and y with macros

PR #194, opened 2016-12-02
[WIP] tentative optional implementation

piiswrong:
use with 
```
struct SliceParam : public dmlc::Parameter<SliceParam> {
  int axis;
  optional<int> begin;
  optional<int> end;
  DMLC_DECLARE_PARAMETER(SliceParam) {
    DMLC_DECLARE_FIELD(axis).set_lower_bound(0)
      .describe("The axis to be sliced");
    DMLC_DECLARE_FIELD(begin)
      .set_default(optional<int>())
      .describe("The beginning index to be sliced");
    DMLC_DECLARE_FIELD(end)
      .set_default(optional<int>())
      .describe("The end index to be sliced");
  }
};

```

@sxjscience @tqchen

### DIFF
--- a/include/dmlc/optional.h
+++ b/include/dmlc/optional.h
@@ -41,7 +41,7 @@ class optional {
     *reinterpret_cast<T*>(&val) = value;
   }
   /*! \brief construct an optional object with another optional object */
-  optional (const optional<T>& other) {
+  optional(const optional<T>& other) {
     is_none = other.is_none;
     if (!is_none) {
       *reinterpret_cast<T*>(&val) =
@@ -87,7 +87,7 @@ class optional {
     return *this;
   }
   /*! \brief non-const dereference operator */
-  T& operator*() {
+  T& operator*() {  // NOLINT(*)
     return *reinterpret_cast<T*>(&val);
   }
   /*! \brief const dereference operator */
@@ -105,11 +105,12 @@ class optional {
   }
   /*! \brief whether this object is holding a value */
   explicit operator bool() const { return !is_none; }
+
  private:
   // whether this is none
   bool is_none;
   // on stack storage of value
-  std::aligned_storage<sizeof(T), sizeof(void*)> val;
+  typename std::aligned_storage<sizeof(T), sizeof(void*)>::type val;
 };
 
 /*! \brief serialize an optional object to string.
@@ -123,7 +124,7 @@ class optional {
  */
 template<typename T>
 std::ostream &operator<<(std::ostream &os, const optional<T> &t) {
-  if (bool(t)) {
+  if (t) {
     os << *t;
   } else {
     os << "None";
@@ -164,6 +165,6 @@ std::istream &operator>>(std::istream &is, optional<T> &t) {
 
 DMLC_DECLARE_TYPE_NAME(optional<int>, "optional<int>");
 
-}  // namespace optional
+}  // namespace dmlc
 
 #endif  // DMLC_OPTIONAL_H_
--- a/include/dmlc/optional.h
+++ b/include/dmlc/optional.h
@@ -27,6 +27,16 @@ class optional {
  public:
   optional() : val(nullptr) {}
   explicit optional(const T& value) : val(new T(value)) {}
+  optional (const optional<T>& other) {
+    if (other) {
+      val = new T(*other);
+    } else {
+      val = nullptr;
+    }
+  }
+  ~optional() {
+    if (val != nullptr) delete val;
+  }
   optional<T>& operator=(const T& value) {
     if (val != nullptr) {
       delete val;
--- a/include/dmlc/optional.h
+++ b/include/dmlc/optional.h
@@ -15,13 +15,14 @@
 #include "./type_traits.h"
 
 namespace dmlc {
-/* \brief dummy type for assign null to optional */
+/*! \brief dummy type for assign null to optional */
 struct nullopt_t {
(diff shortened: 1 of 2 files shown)


tqchen:
I think simply use pair of bool and T is better since the data will sit on stack

piiswrong:
I thought about that. But that will require a default constructor. Do you think its going to be a problem

tqchen:
take a look at the implementation of any. You can have a aligned storage that can be used to initialize T

tqchen:
https://github.com/dmlc/dmlc-core/blob/master/include/dmlc/any.h#L139 see how to use new and delete on the stack storage

piiswrong:
done

include/dmlc/optional.h
```diff
@@ -0,0 +1,169 @@
...
+    }
+    return *reinterpret_cast<const T*>(&val);
+  }
+  /*! \brief whether this object is holding a value */
+  explicit operator bool() const { return !is_none; }
+ private:
+  // whether this is none
+  bool is_none;
+  // on stack storage of value
+  std::aligned_storage<sizeof(T), sizeof(void*)> val;
```
tqchen: chang sizeof(void*) to alignof(T)

tqchen: use under score for class member

include/dmlc/optional.h
```diff
@@ -0,0 +1,169 @@
...
+   */
+  optional<T>& operator=(nullopt_t) {
+    if (!is_none) {
+      reinterpret_cast<T*>(&val)->~T();
+      is_none = true;
+    }
+    return *this;
+  }
+  /*! \brief non-const dereference operator */
+  T& operator*() {
```
tqchen: nolint

tqchen:
added a few more comments. Please add a testcase to https://github.com/dmlc/dmlc-core/tree/master/test

tqchen:
seems there lacks move constructor, and move assignment, consider use copy swap idiom https://en.wikibooks.org/wiki/More_C%2B%2B_Idioms/Copy-and-swap

c.f. https://github.com/dmlc/dmlc-core/blob/master/include/dmlc/any.h#L229

tqchen:
please act on the comments let let me know when I can do another round of review

include/dmlc/optional.h
```diff
@@ -0,0 +1,170 @@
...
+ */
+template<typename T>
+class optional {
+ public:
+  /*! \brief construct an optional object that contains no value */
+  optional() : is_none(true) {}
+  /*! \brief construct an optional object with value */
+  explicit optional(const T& value) {
+    is_none = false;
+    *reinterpret_cast<T*>(&val) = value;
```
tqchen: This is wrong when value have non-trivial destructor (destructor will get invoked on val, which is not T). Use the following code instead
```c++
new ((&val)  T(value)
```

include/dmlc/optional.h
```diff
@@ -0,0 +1,170 @@
...
+  }
+  /*! \brief construct an optional object with another optional object */
+  optional(const optional<T>& other) {
+    is_none = other.is_none;
+    if (!is_none) {
+      *reinterpret_cast<T*>(&val) =
```
tqchen: same as the previous comment

include/dmlc/optional.h
```diff
@@ -0,0 +1,170 @@
...
+    is_none = false;
+    *reinterpret_cast<T*>(&val) = value;
+    return *this;
+  }
+  /*! \brief set this object to hold the same value with other
+   *  \param other the other object
+   */
+  optional<T>& operator=(const optional<T> &other) {
+    if (!is_none) {
+      reinterpret_cast<T*>(&val)->~T();
```
tqchen: implement a swap function between optional<T>, and simply use

```c++
(optional<T>(other)).swap(*this)
```

include/dmlc/optional.h
```diff
@@ -0,0 +1,170 @@
...
+    return *reinterpret_cast<const T*>(&val);
+  }
+  /*! \brief whether this object is holding a value */
+  explicit operator bool() const { return !is_none; }
+
+ private:
+  // whether this is none
+  bool is_none;
+  // on stack storage of value
+  typename std::aligned_storage<sizeof(T), sizeof(void*)>::type val;
```
tqchen: change 

```
sizeof(void*) to  alignof(T)
```
Otherwise the alignment can be wrong for some types.

include/dmlc/optional.h
piiswrong: why would destructor be invoked?

tqchen:
Still have a few problems, I summarize them here

- Use of assignment operator= in constructor, instead of new operator(copy/move constructor)
   - This causes problem, because the assignment assumes that the content is a valid T. Imagine we have a T that originally have a internal pointer, and deletes the old pointer when it get assigned,  this will results in a segfault, because the real T is not in correct state.

- Alignment specification, need to be 
   ```c++
     std::aligned_storage<sizeof(T), alignof(T)>
   ```
   - In any it is sizeof(void*) alignment, but notice the choosing template to only choose T in stack of the alignment matches.

- Implementation of copy assignment operator. The swap idiom could be useful, because it reuses the destructor logic and constructor to implement the copy, so there is no need to put destruction logic inside assignment

tqchen:
The current code is OK, given the comments are fixed. It might be also good to add a move constructor and move assignment, to make rule of five

PR #197, opened 2016-12-05
Add more support for CSV

AbdealiJK:


include/dmlc/parameter.h
```diff
@@ -853,4 +853,78 @@ inline ValueType GetEnv(const char *key,
   return ret;
 }
 }  // namespace dmlc
+
+// define string serializer for vector, to get the arguments
+namespace std {
```
AbdealiJK: I tried adding this into a new
```
template<>
class FieldEntry<std::vector<int> >
    : public FieldEntryBase<FieldEntry<std::vector<int> >, std::vector<int> > {
```
but for some reason, it kept ignoring this and ran https://github.com/dmlc/dmlc-core/pull/197/commits/1163f51ee47feb370529184e363e407574fe93a1#diff-4b3c194d1c3981ab9fe7006fe3428563R518 instead. I have absolutely no idea why that generic template was taking higher priority than this one.

PS: This is taken from the param.h in XGBoost

tqchen:
for the set structure, we could consider a sorted array and use std::binary_search instead. I am traveling and may not be able to act on this too soon

AbdealiJK:
The set structure uses a tree internally and is logN

src/data/csv_parser.h
```diff
@@ -76,18 +107,31 @@ ParseBlock(char *begin,
     char* p = lbegin;
     int column_index = 0;
     IndexType idx = 0;
-    float label = 0.0f;
+    float label = NAN;
```
AbdealiJK: Hm, I see. I personally found it weird - but will do

tqchen:
Yes, but the tree structure might have worse cache locality than binary search on vector(due to the separate allocation of node)

AbdealiJK:
Hm, I see and that makes sense.
Thinking in hind sight I think a better idea may be to use a unordered_set - (uses a hash function)
Or if you think that the hash function may not be ideal, I'd rather make a dense array with 0s and 1s as the number of features would be relatively small but the query would happen num_row times. (On most datasets)

i.e. the memory overhead of `num_col` is better than the cost of computing `num_row * log(num_col)`

tqchen:
dense array is the right way to go if there are not a lot of features, but since dmlc-core need to be compatible to these cases.

Having a dense array, and make it size=0 when there is no ignore is an good option to go though

AbdealiJK:
It seems the linux tests fail even if I do not add my modifications. So, I wonder if it ever worked in Linux...

PR #199, opened 2016-12-14
Minor make adjustments, fix warnings

cjolivier01:


Makefile
```diff
@@ -79,7 +73,7 @@ data.o: src/data.cc
 recordio.o: src/recordio.cc
 config.o: src/config.cc
 
-$(LIB_DIR)/libdmlc.a: $(OBJ)
```
piiswrong: why removing these?

src/io/s3_filesys.cc
```diff
@@ -824,10 +824,10 @@ void ListObjects(const URI &path,
 S3FileSystem::S3FileSystem() {
   const char *keyid = getenv("AWS_ACCESS_KEY_ID");
   const char *seckey = getenv("AWS_SECRET_ACCESS_KEY");
-  if (keyid == NULL) {
+  if (keyid == NULL || !*keyid) {
     LOG(FATAL) << "Need to set enviroment variable AWS_ACCESS_KEY_ID to use S3";
   }
-  if (seckey == NULL) {
+  if (seckey == NULL || !*seckey) {
```
piiswrong: @mli Is this correct?

piiswrong: @mli is this correct?

cjolivier01: My implementation returns NULL (Ubuntu). However, the documentation is not specific on this and states that it returns NULL if the environment variable is not found.  However, I can put an empty variable in the environment:
[coolivie@DEV:~]export FOO=
[coolivie@DEV:~]env | grep FOO
FOO=
[coolivie@DEV:~]

cjolivier01: By googling that, I see cases of some implementations returning an empty string.

Also, this example does not appear to check:  
http://cvsweb.netbsd.org/bsdweb.cgi/src/lib/libc/stdlib/getenv.c?rev=1.19&content-type=text/x-cvsweb-markup

piiswrong: ok I'll merge this since it looks like it wouldn't hurt.

PR #200, opened 2016-12-15
Add std::vector for params and add ignore_columns in LibSVM

AbdealiJK:


AbdealiJK:
@tqchen Was finally able to fix the Travis issue. It turned out to be a bug in TextParser when multiple threads are being used but a param to use only limited threads is given - Check https://github.com/dmlc/dmlc-core/pull/200/commits/f1d60a7307f6b65db5bde9a0eb951299ed561cda

Could we get this merged ?

PR #201, opened 2016-12-16
add_enum for optional

piiswrong:
@tqchen


tqchen:
LGTM, but please add a unittest case before merge

piiswrong:
@tqchen added test

PR #202, opened 2016-12-18
add Mac support, fix issue #645 on `mxnet`;

liangfu:
add Mac support, and trying to fix issue on `mxnet`, see https://github.com/dmlc/mxnet/issues/645 .


tqchen:
Before merge this, we would like to confirm if it depends on availability of openmp enabled compiler, since clang by default does not yet come with it, and forcing this option can cause problem

liangfu:
Latest clang version brought by XCode actually support openmp, one can enable the feature by checking `-Xclang -fopenmp` flags, however the openmp runtime library is missing from the default platform, that is to say the the openmp runtime library file `libiomp5.dylib` is not available, as mentioned at [OpenMP/Clang](https://clang-omp.github.io) . And the library can be downloaded from [Intel OpenMP Runtime Library](http://openmp.llvm.org/).

tqchen:
So we will need to make sure that the cmake pass compile when openmp is not available and openmp option is switched off, the current flag seems to force it even when openmp runtime is not available

PR #203, opened 2016-12-20
Use path style in S3 FileSystem, to avoid the issue of period in buck…

DamonDeng:
Use path style in S3 File System in dmlc-core/src/io/s3_filesys.h and dmlc-core/src/io/s3_filesys.cc

The virtual host style Restful request of s3 File System will cause the SSL issue when there is a period in bucket name.

The pull request use path style Restful request to resolve that issue. The path style request requires information of the Region in which the S3 bucket was created. 
End user can set the environment `AWS_REGION` to the region they want, such as `us-east-1`, `ap-southeast-1`.

For backward compatibility, if no `AWS_REGION` was set and there is no period in the bucket name, this fix will still use original virtual host style Restful request.

This fix was tested in `us-east-1` and `ap-southeast-1`.

For reminding , S3 FileSystem in this project is using V2 signature of S3, which can't support new region like Frankfurt. 

We need to use V4 signature of S3 in the future.

### DIFF
--- a/src/io/s3_filesys.cc
+++ b/src/io/s3_filesys.cc
@@ -178,14 +178,14 @@ size_t WriteStringCallback(char *buf, size_t size, size_t count, void *fp) {
   return size;
 }
 
-std::string getEndpoint(std::string region_name){
+std::string getEndpoint(std::string region_name) {
   // using if elseif chain switching region_name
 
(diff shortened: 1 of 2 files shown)


piiswrong:
@mli

PR #204, opened 2016-12-30
fix complie error with vs2013

yajiedesign:
fix complie error with vs2013
fix lint with windows


scripts/lint.py
```diff
@@ -121,7 +121,7 @@ def get_header_guard_dmlc(filename):
         file_path_from_root = _HELPER.project_name +  file_path_from_root[idx + 3:]
     else:
         for spath in inc_list:
-            prefix = spath + os.sep
+            prefix = spath + '/'
```
piiswrong: why?

yajiedesign: because  in windows,this path also use '/'

include/dmlc/optional.h
```diff
@@ -15,10 +15,16 @@
 #include "./type_traits.h"
 
 namespace dmlc {
+
 /*! \brief dummy type for assign null to optional */
 struct nullopt_t {
+#if defined(_MSC_VER) && _MSC_VER < 1900
+  /*! \brief dummy constructor */
+  explicit nullopt_t(int) {}
+#elif
```
piiswrong: did you mean else?

yajiedesign: yes....strange, I remember I writing else

PR #220, opened 2017-02-11
add DMLC_SERVER_ID and DMLC_WORKER_ID to enable Map-Reduce on Workers

jli05:
Add environment variables `DMLC_SERVER_ID` and `DMLC_WORKER_ID` before launching code via `ssh` in order to realise Map-Reduce operations on Workers.

### DIFF
--- a/tracker/dmlc_tracker/ssh.py
+++ b/tracker/dmlc_tracker/ssh.py
@@ -70,6 +70,10 @@ def submit(args):
         # launch jobs
         for i in range(nworker + nserver):
             pass_envs['DMLC_ROLE'] = 'server' if i < nserver else 'worker'
+            if i < nserver:
+                pass_envs['DMLC_SERVER_ID'] = i
+            else:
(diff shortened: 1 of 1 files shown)


jli05:
We're doing a project for Prof Anima Anandkumar to implement the Spectral LDA algorithm for MXNet, which starts with Map-Reduce operations on the word count vectors of the document collection. Could we push this PR into `dmlc-core` so that we could proceed?

jli05:
Actually could we revert the merge? I noticed there're fields `mxnet.KVStore.num_workers` and `mxnet.KVStore.rank`. They should suffice for running MapReduce operations.

tqchen:
can you pass a revert PR?

jli05:
I think you may be able to directly do it from here. Let me know otherwise.

https://help.github.com/articles/reverting-a-pull-request/

tqchen:
I cannot directly revert it maybe due to followup changes, please submit a PR to do so

PR #233, opened 2017-03-16
Build static library in dmlc-core

tqchen:


tqchen:
@cjolivier01 do you have a specific reason to specify this as shared before? Since ideally this should be directly linked into the project that uses it

tqchen:
Specifically, specifying shared seems make windows build fail because dll is build instead of .lib file

PR #236, opened 2017-04-13
[master] Ignore warning "-Wplacement-new".

hotpxl:
When compiling MXNet, this use of `any` generates a lot of warnings. This `pragma` temporarily disables this warning for GCC.

### DIFF
--- a/include/dmlc/any.h
+++ b/include/dmlc/any.h
@@ -182,7 +182,10 @@ inline any::any(T&& other) {
                   "Any can only hold value that is copy constructable");
     type_ = TypeInfo<DT>::get_type();
     if (data_on_stack<DT>::value) {
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wplacement-new"
       new (&(data_.stack)) DT(std::forward<T>(other));
+#pragma GCC diagnostic pop
     } else {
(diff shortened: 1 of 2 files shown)


hotpxl:
@piiswrong Any comment?

hotpxl:
I use GCC 6 and I get screens of warnings compiling MXNet. I found two major sources. There are some others as well in operator implementation, mainly comparing signedness. I did not bother to correct those.

tqchen:
One thing we can try is to switch to experimental/any when it is avaliable (c++14)

sxjscience:
I find this PR causes the warning here http://ec2-52-25-96-65.us-west-2.compute.amazonaws.com/blue/organizations/jenkins/mxnet/detail/PR-5946/1/pipeline

```
/workspace/dmlc-core/include/dmlc/././any.h:186:32: error: unknown option after '#pragma GCC diagnostic' kind [-Werror=pragmas]
```

tqchen:
hmm, @hotpxl mind push a followup that does framework detection? Or cplusplus_version detection to switch to experimental/any.h?

hotpxl:
looks like older version of gcc does not have this error and thus cannot disable it. i'll see what i can do

PR #239, opened 2017-04-20
local job submit once

formath:
Before fixed:
1. If job submit error, the job will keep trying to submit again and again so can't be killed.
2. When developing an async training application based on ps-lite and dmlc-core,  an error occur when a node finalize after the whole job finished, this node process will be submitted again and send a ADD_NODE message even training have been done.
3. It is meaningless to re-submit a process when an error occur not in the beginning of the job because most job are stateful. In this circumstance, to re-submit the whole job rather than just re-submit one of the processes will be more reasonable.

After fixed:
All local job will submit once. If an error occur, user could get the error immediately.
@tqchen


tqchen:
can we add a flag to change between keep alive option and kill, maybe --num-retry

formath:
use DMLC_NUM_ATTEMPT as retry_num

PR #242, opened 2017-04-23
[master] -Wplacement-new is introduced in GCC 6.

hotpxl:
@sxjscience Could you please check whether this fixes the compilation error?

I was not aware that "-Wplacement-new" does not exist in previous versions.


sxjscience:
I've checked. It has solved the problem.

PR #243, opened 2017-04-25
Disable 4068 unknown pragma warning in VS

sxjscience:
Disable the warning in Visual Studio.


sxjscience:
The VS reports lots of warnings due to lines like this https://github.com/dmlc/dmlc-core/blob/master/include/dmlc/any.h#L185 .

Issue #215, opened 2017-01-27
Using temporary credentials (Instance profiles) requires session token.


sandeep-krishnamurthy:
If a user wants to use AWS temporary credentials provided via instance profiles  then user needs to send a session token that has to go to requests to AWS: http://docs.aws.amazon.com/AmazonS3/latest/dev/RESTAuthentication.html#UsingTemporarySecurityCredentials​ 
in addition to the access key and secret key.

ryanjoneil:
If I understand correctly, PR #254 should handle this.

PR #254, opened 2017-05-17
Adds AWS_SECURITY_TOKEN environment variable to S3FileSystem HTTP requests

ryanjoneil:
Due to security policies, some organizations won't let you set access and secret keys in environment variables on EC2 instances. However, if IAM policies are set then it is possible to get keys from instance metadata:

http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html#instance-metadata-security-credentials

One can `curl` the URL in those docs for the appropriate security credentials and then call `xgboost` as before.

Using temporary credentials requires the session token be set on the signature and HTTP headers. This PR adds that capability through an `AWS_SESSION_TOKEN` environment variable.

### DIFF
--- a/src/io/s3_filesys.cc
+++ b/src/io/s3_filesys.cc
@@ -465,10 +465,10 @@ class ReadStream : public CURLReadStreamBase {
   ReadStream(const URI &path,
              const std::string &aws_id,
              const std::string &aws_key,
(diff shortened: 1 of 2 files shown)


ryanjoneil:
If I understand correctly, this should fix #215.

PR #257, opened 2017-05-28
Use GNUInstallDirs for standard conforming install target.

de-vri-es:
This PR replaces the custom commands for install targets with the CMake `install()` command. Additionally, it uses `GNUInstallDirs.cmake` [1] to install to different types of files to the right location.

Additionally, `GNUInstallDirs` allows the user to override the install location of different types of files with well-known variable names (`-DCMAKE_INSTALL_LIBDIR`, `-DCMAKE_INSTALL_INCLUDEDIR`, `...`).

This makes it much easier to package the software for a package manager.

[1] https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -131,9 +131,13 @@ if(INSTALL_INCLUDE_DIR)
 endif()
 
 # ---[ Install the archive static lib and header files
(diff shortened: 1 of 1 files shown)


hcho3:
Closing this in favor of #495.

PR #259, opened 2017-05-30
Update Row class so that it's easier to construct tensors for label and weights in MXNet

eric-haibin-lin:
@tqchen this change is mainly for https://github.com/eric-haibin-lin/mxnet/pull/55 so that I don't have to allocate extra memory to hold the label in libsvm iter. Will changing interface introduce problems for other projects depending on dmlc-core?


tqchen:
LGTM

PR #262, opened 2017-06-05
Link with libhdfs statically

superbobry:
This way we don't need libhdfs.so in the JAR for xgboost4j.


tqchen:
hmm, seems i merged that PR to quickly @superbobry can you also fix that?  Just break that macro into two lines

superbobry:
Done, sorry for not honoring the linter rules.

tqchen:
https://travis-ci.org/dmlc/dmlc-core/jobs/239892100 You can reproduce linter check locally by type make lint

superbobry:
I've updated the PR to merge cleanly with `master`.

superbobry:
I've updated the PR again.

byronyi:
Why this has been merged? If `libhdfs.a` isn't compiled with `-fPIC`, it would be impossible to build a shared library, e.g. `libmxnet.so` when using `USE_HDFS=1`.

superbobry:
Ouch, I was not aware of that at the time.

How common is it to compile `libhdfs` without `-fPIC`? Shared linking of `libhdfs` did complicate `xgboost4j` maintenance quite a bit, therefore this PR.

byronyi:
It seems that’s an issue with various versions of CDH...instead of forcing the build with either a static lib or a dynamic one, how about adding a knob flag to control this behavior?

hcho3:
@byronyi Would you be able to file a pull request to add an option?

thvasilo:
@byronyi I'm running into this issue when trying to build XGBoost with HDFS support. Is there some way to ensure libhdfs is built with `-fPIC`?

PR #282, opened 2017-07-14
Added support for optional<bool> parameters.

DickJC123:
The following adds the hooks to permit optional bool parameters, used as in:
  dmlc::optional&lt;bool&gt; my_bool;
  ...
  DMLC_DECLARE_FIELD(my_bool)
        .set_default(dmlc::optional&lt;bool&gt;())
        .describe("My optional bool parameter.  User can set to 0/1/true/false, defaults to 'None'");

### DIFF
--- a/include/dmlc/optional.h
+++ b/include/dmlc/optional.h
@@ -217,6 +217,7 @@ inline std::istream &operator>>(std::istream &is, optional<bool> &t) {
 
 /*! \brief description for optional int */
 DMLC_DECLARE_TYPE_NAME(optional<int>, "int or None");
+/*! \brief description for optional bool */
 DMLC_DECLARE_TYPE_NAME(optional<bool>, "boolean or None");
 
 }  // namespace dmlc
--- a/include/dmlc/optional.h
+++ b/include/dmlc/optional.h
@@ -7,6 +7,7 @@
 #define DMLC_OPTIONAL_H_
 
 #include <iostream>
+#include <string>
 #include <utility>
 #include <algorithm>
 
@@ -107,6 +108,8 @@ class optional {
   }
   /*! \brief whether this object is holding a value */
   explicit operator bool() const { return !is_none; }
+  /*! \brief alternate, more-explicit check of object value */
+  const bool IsNone() const { return is_none; }
 
  private:
   // whether this is none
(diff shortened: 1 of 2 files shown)


include/dmlc/optional.h
```diff
@@ -107,6 +108,8 @@ class optional {
   }
   /*! \brief whether this object is holding a value */
   explicit operator bool() const { return !is_none; }
+  /*! \brief alternate, more-explicit check of object value */
+  const bool IsNone() const { return is_none; }
```
piiswrong: Is this interface conform to c++17 standard?

include/dmlc/optional.h
```diff
@@ -173,9 +176,49 @@ std::istream &operator>>(std::istream &is, optional<T> &t) {
...
+ *    s2 >> x;  // x == optional<bool>()
+ *  \endcode
+ *
+ *  \param is input stream
+ *  \param t target optional<bool> object
+ *  \return input stream
+ */
+inline std::istream &operator>>(std::istream &is, optional<bool> &t) {
+  std::string s;
+  is >> s;
```
piiswrong: Could you check if reading everything in works?
If you have Tuple<optional<bool> > would this read subsequent elements?

DickJC123: I can't think of a Tuple-related use that would be a problem.  Feel free to suggest one.  First of all,  optional&lt;bool&gt; are not initialized by strings.  There's a no-arg constructor and one that takes a bool reference.  The operator>> implementation is used to set an already-constructed instance.  Although this isn't exactly your concern, I've added a test to prove the following works:

    dmlc::optional&lt;bool&gt;  x, y;
    std::istringstream is("false true");
    is >> x >> y;
    CHECK_EQ(x.value(), 0);
    CHECK_EQ(y.value(), 1);

I've also verified that in the Python context, if a user does  mx.symbol.SomeOperator(...,my_opt_bool_param="false more_junk",...)

...then an error is generated.

If you're tempted to look into this further, keep in mind that the new operator>> specialization is designed to be invoked by the following code within parameter.h:
template<typename TEntry, typename DType>
class FieldEntryBase : public FieldAccessEntry {
  virtual void Set(void *head, const std::string &value) const {
    std::istringstream is(value);
    **is >> this->Get(head);**

Feel free to describe your Tuple scenario more specifically and I'll investigate further.

include/dmlc/optional.h
DickJC123: I removed the pointless initial 'const' keyword- there by mistake.  Is this what you were referring to?

DickJC123: After some research, I see that the existing code includes an explicit conversion from an optional&lt;T&gt; to bool.  Since C++11, these can be considered in many common "contextual conversions to bool," including the controlling expression of if-statements.  Note the absence of 'operator' in the IsNone() method I'm proposing, so I don't see any issues in disturbing the existing conversion behavior.

My motivation in adding IsNone() is one of coding clarity.  In the case of optional&lt;int&gt; , the coder is often protected from leaving off a .value() call (which returns an int, so different from bool).  In the case of optional&lt;bool&gt;, no such protection exists.  I was hoping to steer the usage to the more explicit (and in my mind more understandable):

if ( my_optional_bool.IsNone() )
    my_optional_bool = some_good_default_value;

Without the IsNone() method, this must appear as:

if ( !my_optional_bool )
    my_optional_bool = some_good_default_value;

Someone not intimately familiar with the optional&lt;T&gt; class might think "if my_optional_bool is false, then set it to some_good_default_value."

Bottom line, I can remove this from the PR if it's getting in the way of its acceptance.  It would be my preference to keep it, but it represents no fundamental new functionality.

piiswrong: dmlc::optional is a temporary replacement for std:optional until we move to c++17.
The goal is to maintain consistency with std::optional interface.

include/dmlc/optional.h
piiswrong: Could you check if this works?

```
nnvm::Tuple<dmlc::optional<bool> > x;
std::istringstream is("(false, true, false, false)");
is >> x;
```

DickJC123: I've fixed the istream reader to pass on your posed nnvm::Tuple testcase.  I can contribute a completed version of the test to nnvm after this dmlc-core PR is accepted, if you wish.

include/dmlc/optional.h
DickJC123: I've dropped IsNone() and added the C++17 standard's has_value(), which satisfies my desire for a more-explicit call.

PR #284, opened 2017-07-18
Reintroducing threaded prefetching to reader

ptrendx:
With the improved IO pipeline in MXNet reading from disk started to become quite big issue. That is why this commit reintroduces the threaded prefetching to InputSplit, while at the same time setting max capacity to 2 in order to conserve memory. In my test I did not really see any difference in memory consumption between current and proposed version, while IO speed improved by ~20% (480px images, quality 95, random resize, crop, mirror for augmentations).
Having IO prefetched also makes it possible to use more time for it to e.g. perform better shuffling (cost of random seek is hidden by the rest of the IO pipeline).

### DIFF
--- a/src/io.cc
+++ b/src/io.cc
@@ -11,6 +11,7 @@
 #include "io/filesys.h"
 #include "io/local_filesys.h"
 #include "io/cached_input_split.h"
+#include "io/threaded_input_split.h"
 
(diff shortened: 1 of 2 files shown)


piiswrong:
@tqchen

PR #289, opened 2017-07-26
[WIP] Indexed recordIO support in InputSplit

ptrendx:
Not complete yet, don't merge

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,7 @@ endif
 
 .PHONY: clean all test lint doc example pylint
 
-OBJ=line_split.o recordio_split.o input_split_base.o io.o local_filesys.o data.o recordio.o config.o
+OBJ=line_split.o indexed_recordio_split.o recordio_split.o input_split_base.o io.o local_filesys.o data.o recordio.o config.o
 
 ifeq ($(USE_HDFS), 1)
 	OBJ += hdfs_filesys.o
@@ -63,6 +63,7 @@ example: $(ALL_EXAMPLE)
 
 line_split.o: src/io/line_split.cc
 recordio_split.o: src/io/recordio_split.cc
+indexed_recordio_split.o: src/io/indexed_recordio_split.cc
 input_split_base.o: src/io/input_split_base.cc
 hdfs_filesys.o: src/io/hdfs_filesys.cc
 s3_filesys.o: src/io/s3_filesys.cc
--- a/include/dmlc/io.h
+++ b/include/dmlc/io.h
@@ -263,7 +263,8 @@ class InputSplit {
                             const char *index_uri,
                             unsigned part_index,
                             unsigned num_parts,
-                            const char *type);
+                            const char *type,
+                            const size_t batch_size = 256);
 };
 
 /*!
--- a/include/dmlc/io.h
+++ b/include/dmlc/io.h
@@ -210,11 +210,6 @@ class InputSplit {
   virtual bool NextBatch(Blob *out_chunk, size_t n_records) {
     return NextChunk(out_chunk);
   }
-  virtual void SetShuffle(bool shuffle) {
-    if (shuffle) {
-      LOG(WARNING) << "shuffle option not supported by this InputSplit, ignoring";
-    }
-  }
   /*! \brief destructor*/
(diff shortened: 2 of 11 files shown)


piiswrong:
@tqchen

include/dmlc/io.h
```diff
@@ -205,19 +227,56 @@ class InputSplit {
...
+   * \return a new input split
+   * \sa InputSplit::Type
+   */
+  static InputSplit* Create(const char *uri,
+                            const char *index_uri,
+                            unsigned part_index,
+                            unsigned num_parts,
+                            const char *type,
+                            const bool shuffle = false,
+                            const int seed = 0,
```
piiswrong: not sure defaulting seed to 0 is a good idea.

include/dmlc/io.h
```diff
@@ -205,19 +227,56 @@ class InputSplit {
...
+                            const size_t batch_size = 256);
```
piiswrong: batch_size shouldn't have a default value

src/io/cached_input_split.h
```diff
@@ -88,7 +88,7 @@ class CachedInputSplit : public InputSplit {
     LOG(FATAL) << "ResetPartition is not supported in CachedInputSplit";
   }
   virtual void HintChunkSize(size_t chunk_size) {
-    buffer_size_ = std::max(chunk_size / sizeof(size_t), buffer_size_);
+    buffer_size_ = std::max(chunk_size / sizeof(uint32_t), buffer_size_);
```
piiswrong: ?

ptrendx: Even though recordio is aligned to 4 bytes, reader was reading to chunk's data vector which was size_t (so 8 bytes aligned). This is problematic, because there is no guarantee that recordio entry is aligned to 8 bytes. That's why I changed chunk's internal data vector to uint32_t

include/dmlc/io.h
ptrendx: It is added to magic number inside indexedrecordioinputsplit. It is actually the second as in image iterator in MxNet (seed defaults to 0 + magic).

include/dmlc/io.h
ptrendx: This is only a hint (like the HintChunkSize) and only IndexedRecordioInputSplit actually uses that. Readers can give back any number of samples anyway so consumers of the read data need to be prepared to handle that.

tqchen:
can you comment a bit about index file format and mechanism, e.g. does every shard need to load the entire index file, etc,

ptrendx:
It is very simple file format (line per entry with 2 integers - entry id and offset) introduced in https://github.com/apache/incubator-mxnet/pull/2887 a year ago, but not actually used anywhere (although im2rec.py script actually generates the index file alongside recordIO file).
Each worker needs the whole index file, then each worker is responsible for only its own portion of the indexes.
Shuffle introduces global shuffling, but only inside a worker part (so that there is no need to share seeds between workers etc.). I thought about making it so that groups of few entries are shuffled in case the performance is bad, but at least on systems I tested it on it was fine as-is. This should really improve the shuffle option in MXNet recordio iterator, because right now it shuffles only inside a single chunk, which often is less than or comparable to single batch (so there is effectively no shuffling).
Motivation for this PR is twofold:
 - I introduced new much faster IO pipeline in MXNet (https://github.com/apache/incubator-mxnet/pull/7152) which unfortunately makes it hard to implement shuffling inside iterator, so shuffling needs to be done on lower level.
 - having InputSplit return precisely the batch size of images makes the IO pipeline even faster, because the overflow storage is not used, so there are no unnecessary copies being made, and also the cost of OpenMP synchronization between threads is paid only once per batch (In my tests on DGX-1 I went from 6.2k imgs/s to 7k imgs/s when testing just IO on RN50 pipeline).

Makefile
```diff
@@ -63,6 +63,7 @@ example: $(ALL_EXAMPLE)
 
 line_split.o: src/io/line_split.cc
 recordio_split.o: src/io/recordio_split.cc
+indexed_recordio_split.o: src/io/indexed_recordio_split.cc
```
piiswrong: may need to add this to amalgamation

PR #292, opened 2017-08-04
add cpplint logic for handling plugin

szha:
this is to fix cpplint


szha:
When I run cpplint locally on my mac, it complains about all the plugin headers having the wrong header guard. It might be specific to my environment, and I was able to fix the cpplint in mxnet without this change. Closing for now.

PR #294, opened 2017-08-08
[PARAMETER] Support init and update default back to kwargs

tqchen:


tqchen:
@piiswrong @ZihengJiang 

We can update param_parser to use this, so kwargs relfects the default value as well

PR #299, opened 2017-08-22
add hash and equal for optional

piiswrong:


include/dmlc/common.h
```diff
@@ -26,6 +26,24 @@ inline std::vector<std::string> Split(const std::string& s, char delim) {
...
+ * \brief hash an object and combines the key with previous keys
+ */
+template<typename T>
+inline size_t HashCombine(size_t key, const T& value) {
+  std::hash<T> hash_func;
+  return key ^ (hash_func(value) + 0x9e3779b9 + (key << 6) + (key >> 2));
+}
+
+/*!
+ * \brief specialize for size_t
```
larroy: Why would you specialize for size_t?  

Also the comment is not very good as it's obvious the what, but not the why.

PR #300, opened 2017-08-26
don't turn on -fopenmp on osx

szha:
clang on osx doesn't support openmp option out of the box.


szha:
The doc/Doxyfile was updated using `doxygen -u` as suggested by the build errors.

PR #315, opened 2017-09-25
Small param bug

KellenSunderland:
This fixes a small issue I've run across when using a very small learning rate.  When param values are out of range for the stof or stod functions we're throwing an out of range exception and bypassing dmlc::ParamError handling.  

I've also enabled builds for the unit test suite for dmlc-core from cmake with this PR.  I've modeled most of those changes on the CMake files from mxnet.

Testing done: make lint, make doc, make test, cmake, ./dmlc_unittest
@larroy


larroy:
LGTM

PR #321, opened 2017-10-18
Shuffle in groups for IndexedRecordIO

ptrendx:
@zhreshold

This PR batches images when shuffling images with Indexed RecordIO input split to increase performance of reading from disk. The size of the group is set to 16 by default (tested on RN50, with that group size spinning drive was not a bottleneck for MXNet on 6-core CPU doing simple input pipeline of just decoding, cropping and mirroring) and controlled by new environment variable DMLC_INDEXED_RECORDIO_SHUFFLE_GROUP_SIZE.

### DIFF
--- a/src/io/indexed_recordio_split.h
+++ b/src/io/indexed_recordio_split.h
@@ -33,7 +33,7 @@ class IndexedRecordIOSplitter : public InputSplitBase {
                           const int seed = 0) {
     this->shuffle_ = shuffle;
     if (shuffle) SetRandomSeed(seed);
-    this->shuffle_group_size_ = dmlc::GetEnv("DMLC_INDEXED_RECORDIO_SHUFFLE_GROUP_SIZE", 1);
+    this->shuffle_group_size_ = dmlc::GetEnv("DMLC_INDEXED_RECORDIO_SHUFFLE_GROUP_SIZE", 16);
     this->batch_size_ = batch_size;
     this->Init(fs, uri, INDEXED_RECORDIO_ALIGN);
     this->ReadIndexFile(fs, index_uri);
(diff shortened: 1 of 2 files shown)


src/io/indexed_recordio_split.h
```diff
@@ -32,6 +33,7 @@ class IndexedRecordIOSplitter : public InputSplitBase {
                           const int seed = 0) {
     this->shuffle_ = shuffle;
     if (shuffle) SetRandomSeed(seed);
+    this->shuffle_group_size_ = dmlc::GetEnv("DMLC_INDEXED_RECORDIO_SHUFFLE_GROUP_SIZE", 16);
```
zhreshold: Can you add some explanation to docs/faq/env_var.md, people won't know it otherwise

ptrendx: It is in dmlc-core so I think it should start with DMLC_, no? I will add the explanation once I make PR to MXNet with updated dmlc-core.

zhreshold: lol, my bad. didn't notice that

ptrendx:
I don't think it would make big difference (and user is free to change it to lower value if they want) - the indexed recordio version of shuffle is much "stronger" shuffling than with non-indexed recordio. In the non-indexed case you shuffle images inside chunks of 64 MB after my latest change to MXNet, and 8 MB before that change. 8 MB is ~100 images, so if you had batch size bigger than that you were effectively doing no shuffling at all. So if the training accuracy was ok with non-indexed recordio before, it will not be worse now, even with group size of 16.

ptrendx:
@piiswrong @zhreshold Do you have any other comments?

zhreshold:
I'm good with the modification.

tqchen:
@piiswrong  merge this?

hcho3:
Closing, as the pull request has been stable. Feel free to re-submit if the change is still necessary. Thanks!

PR #324, opened 2017-11-01
Pass additional environment variables when launching a job

rahul003:
I found it very useful to pass certain additional variables when launching jobs. I've used this for variables like PYTHONPATH, MXNET_ENGINE_TYPE, MXNET_PROFILER_AUTOSTART, etc. 
I think this would be a good addition to the functionality of launcher. 
If you have suggestions for a different name, please let me know. 
The current usage is 
```--pass-env A,B,C```

I've only tested this with local and ssh. Not tested this with yarn, mpi, sge since I don't have those environments set up. But I think this should work even in those environments. Or should I remove them and only support local and ssh?

Provides a way to solve this issue https://github.com/apache/incubator-mxnet/issues/7857

### DIFF
--- a/tracker/dmlc_tracker/tracker.py
+++ b/tracker/dmlc_tracker/tracker.py
@@ -407,12 +407,20 @@ def get_host_ip(hostIP=None):
     return hostIP
 
 
-def submit(nworker, nserver, fun_submit, hostIP='auto', pscmd=None):
+def submit(nworker, nserver, fun_submit, hostIP='auto', pscmd=None, addnl_envs=None):
     if nserver == 0:
         pscmd = None
 
     envs = {'DMLC_NUM_WORKER' : nworker,
             'DMLC_NUM_SERVER' : nserver}
+
+    if addnl_envs:
+        for k in addnl_envs.split(","):
+            try:
+                envs[k] = os.environ[k]
+            except KeyError:
+                print (k + ' is missing from os environment variables')
+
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: tracker/dmlc_tracker/tracker.py [hunk 1]
```diff
@@ -407,12 +407,17 @@ def get_host_ip(hostIP=None):
...
+def submit(nworker, nserver, fun_submit, hostIP='auto', pscmd=None, addnl_envs=None):
     if nserver == 0:
         pscmd = None
 
     envs = {'DMLC_NUM_WORKER' : nworker,
             'DMLC_NUM_SERVER' : nserver}
+
+    if addnl_envs:
+        for k in addnl_envs.split(","):
+            envs[k] = os.environ[k]
```
anirudh2290: If the key is already present in envs like DMLC_NUM_WORKER or DMLC_NUM_SERVER, shouldn't we skip it ?

rahul003: Well in that case, do you expect the user to have a different value for such environment variables on the host OS, and the value in envs gets changed? 
If they are the same value, it doesn't matter. If they can be different, I can add a check to prevent the user from making a mistake

anirudh2290: Yes I think a check would help here. If there is a env variable set and user has passed it to the script too, we should use the value passed to the script.

PR #341, opened 2017-11-30
GetCurrentThreadId() call needs windows.h

cjolivier01:


tqchen:
why not use http://en.cppreference.com/w/cpp/thread/get_id?

cjolivier01:
This is a 3rd-party optimized lockfree queue. It has some specialized code for different platforms for one reason or another. I am not comfortable changing that code since I assume it's like that for a reason.

tqchen:
OK, I would normally take caution in simply including code into codebase without being able to maintain it. But assuming the moodycamel queue is stable for a while, we might be able to do that, periodically check with upstream is needed

cjolivier01:
ok

cjolivier01:
It'd be better if we could use boost, but that's been outlawed from what I understand :)

tqchen:
as far as I understand, moodycamel queue performs better than boost.lockfree

cjolivier01:
Yeah, they actually don't even use atomics for the simple operations. They sort of work like JEMalloc, where each thread gets its own little section of memory.

tqchen:
Yes, thread local memory management is a common trick for doing things right. We will need to see through on how things can or cannot be improved. While lockfree increases the concurrency, our previous experience find the extra spin sometimes will throttle the CPUs

cjolivier01:
The blocking queue does some spinning, but the reqular, non-blocking queue doesn't. I plan to mostly use the regular one for stuff like pushing profiling data to the profiler (have a PR for that up and some other work going on with it).  I agree the affects of using the blocking queue should be understood and tuned to its specific purpose so as not to starve CPU from other threads.

tqchen:
The non-blocking version requires user to do the spinning actually and need user to poll over multiple queues, a wait mechanism is still needed

cjolivier01:
Depends how you use it.  For DumpProfile(), for instance, the queue is just emptied and then the procedure is over.  There's no reason to block, spin, or otherwise wait for more data.  If you need to wait for more data, then might as well use a blocking queue of some sort.

PR #342, opened 2017-12-02
fix csv parser

eric-haibin-lin:
Fix https://github.com/apache/incubator-mxnet/issues/8587 

@tqchen @anirudh2290  

With the following file content:
```
1,2\n
```

The blocks passed from the multi-threaded [text_parser](https://github.com/dmlc/dmlc-core/blob/master/src/data/text_parser.h#L100) include one block from offset 3 to offset 4. In this case, `lbegin` points to `\n` and should be ignored.

### DIFF
--- a/src/data/csv_parser.h
+++ b/src/data/csv_parser.h
@@ -68,6 +68,7 @@ ParseBlock(char *begin,
   out->Clear();
   char * lbegin = begin;
   char * lend = lbegin;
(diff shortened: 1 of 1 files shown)


eric-haibin-lin:
is the CI not working?
```
$ pip install --user cpplint pylint
/Users/travis/.travis/job_stages: line 57: pip: command not found

```

cjolivier01:
Also not working on my PR for the same reason

PR #346, opened 2017-12-13
Support to allow user to pass specific environment variables (server/worker/both)

rahul003:
Support to allow user to pass specific environment variables (server/worker/both) when using launch.py in MXNet

Example which can be supported now
```--env-server key:value,key2:value2 --env-worker key3:value3,key4:value4 --env key1,key2```

For local and ssh

### DIFF
--- a/tracker/dmlc_tracker/local.py
+++ b/tracker/dmlc_tracker/local.py
@@ -39,7 +39,6 @@ def exec_cmd(cmd, role, taskid, dmlc_envs, addnl_envs):
         env.update(addnl_envs['worker'])
     else:
         env.update(addnl_envs['server'])
-        print(env)
 
     env['DMLC_TASK_ID'] = str(taskid)
     env['DMLC_ROLE'] = role
--- a/tracker/dmlc_tracker/local.py
+++ b/tracker/dmlc_tracker/local.py
@@ -20,10 +20,14 @@ def prepare_envs(args):
     # if --env option was used to pass a variable it would automatically be
     # included because of the next line
     envs['default'] = os.environ.copy()
+
     # given by user
-    for k in args.env:
+    common_envs = [item for item in args.env.split(',') if item]
+    for k in common_envs:
         if k not in envs['default']:
-            raise ValueError('The environment variable '+ k + ' was passed but not set')
+            # raise warning if user intended to set this
+            logging.info('The environment variable %s was passed but not set', k)
+
     envs['server'] = parse_env_pairs(args.env_server)
     envs['worker'] = parse_env_pairs(args.env_worker)
     return envs
--- a/tracker/dmlc_tracker/local.py
+++ b/tracker/dmlc_tracker/local.py
@@ -15,8 +15,10 @@ def prepare_envs(args):
     Load environment variables from arguments
     """
     envs = {}
-    # default env variables passed
-    # would automatically include --envs even if given
+    # envs['default'] refers to the default env variables passed for both worker and server
+
+    # if --env option was used to pass a variable it would automatically be
+    # included because of the next line
     envs['default'] = os.environ.copy()
     # given by user
     for k in args.env:
@@ -76,7 +78,8 @@ def submit(args):
         ----------
         nworker: number of slave process to start up
         nserver: number of server nodes to start up
-        envs: enviroment variables to be added to the starting programs
+        dmlc_envs: refers to the environment variables required to start distributed training
+        addnl_envs: are optional environment variables passed by the user
         """
         procs = {}
         for i in range(nworker + nserver):
--- a/tracker/dmlc_tracker/local.py
+++ b/tracker/dmlc_tracker/local.py
@@ -19,8 +19,11 @@ def prepare_envs(args):
     # would automatically include --envs even if given
     envs['default'] = os.environ.copy()
     # given by user
-    envs['server'] = parse_env_pairs(args.envs_server)
-    envs['worker'] = parse_env_pairs(args.envs_worker)
+    for k in args.env:
+        if k not in envs['default']:
+            raise ValueError('The environment variable '+ k + ' was passed but not set')
+    envs['server'] = parse_env_pairs(args.env_server)
+    envs['worker'] = parse_env_pairs(args.env_worker)
     return envs
 
 def exec_cmd(cmd, role, taskid, dmlc_envs, addnl_envs):
@@ -28,13 +31,15 @@ def exec_cmd(cmd, role, taskid, dmlc_envs, addnl_envs):
     if cmd[0].find('/') == -1 and os.path.exists(cmd[0]) and os.name != 'nt':
         cmd[0] = './' + cmd[0]
     cmd = ' '.join(cmd)
+    env = addnl_envs['default'].copy()
+    for k in dmlc_envs:
+        env[k] = str(dmlc_envs[k])
 
-    env = addnl_envs['default']
-    env.update(dmlc_envs)
     if role == 'worker':
         env.update(addnl_envs['worker'])
     else:
         env.update(addnl_envs['server'])
+        print(env)
 
     env['DMLC_TASK_ID'] = str(taskid)
     env['DMLC_ROLE'] = role
--- a/tracker/dmlc_tracker/local.py
+++ b/tracker/dmlc_tracker/local.py
@@ -8,15 +8,33 @@ import subprocess
 import logging
 from threading import Thread
 from . import tracker
+from .opts import parse_env_pairs
 
-def exec_cmd(cmd, role, taskid, pass_env):
+def prepare_envs(args):
+    """
+    Load environment variables from arguments
+    """
+    envs = {}
+    # default env variables passed
+    # would automatically include --envs even if given
+    envs['default'] = os.environ.copy()
+    # given by user
+    envs['server'] = parse_env_pairs(args.envs_server)
+    envs['worker'] = parse_env_pairs(args.envs_worker)
+    return envs
+
+def exec_cmd(cmd, role, taskid, dmlc_envs, addnl_envs):
     """Execute the command line command."""
     if cmd[0].find('/') == -1 and os.path.exists(cmd[0]) and os.name != 'nt':
         cmd[0] = './' + cmd[0]
     cmd = ' '.join(cmd)
-    env = os.environ.copy()
-    for k, v in pass_env.items():
-        env[k] = str(v)
+
+    env = addnl_envs['default']
+    env.update(dmlc_envs)
+    if role == 'worker':
+        env.update(addnl_envs['worker'])
+    else:
+        env.update(addnl_envs['server'])
 
     env['DMLC_TASK_ID'] = str(taskid)
     env['DMLC_ROLE'] = role
@@ -41,10 +59,11 @@ def exec_cmd(cmd, role, taskid, pass_env):
             else:
                 raise RuntimeError('Get nonzero return code=%d on %s %s' % (ret, cmd, env))
 
-
 def submit(args):
     """Submit function of local jobs."""
-    def mthread_submit(nworker, nserver, envs):
+    envs = prepare_envs(args)
+
+    def mthread_submit(nworker, nserver, dmlc_envs, addnl_envs):
         """
         customized submit script, that submit nslave jobs, each must contain args as parameter
         note this can be a lambda function containing additional parameters in input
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by rahul003** [comment]
@piiswrong Any concerns?

File: tracker/dmlc_tracker/local.py [hunk 1]
```diff
@@ -43,10 +65,11 @@ def exec_cmd(cmd, role, taskid, pass_env):
             else:
                 raise RuntimeError('Get nonzero return code=%d' % ret)
 
-
 def submit(args):
     """Submit function of local jobs."""
-    def mthread_submit(nworker, nserver, envs):
+    envs = prepare_envs(args)
+
+    def mthread_submit(nworker, nserver, dmlc_envs, addnl_envs):
```
anirudh2290: can we change the documentation to include dmlc_envs and addnl_envs ?

tracker/dmlc_tracker/opts.py
```diff
@@ -141,12 +164,14 @@ def get_opts(args=None):
                         directory into remote machines\'s SYNC_DST_DIR')
     parser.add_argument('command', nargs='+',
                         help='Command to be launched')
+
```
anirudh2290: nit: do we need these additional blank lines ?

tracker/dmlc_tracker/ssh.py
```diff
@@ -20,23 +21,67 @@ def sync_dir(local_dir, slave_node, slave_dir):
...
 
-def get_env(pass_envs):
+def prepare_envs(args):
+    """
+    Load environment variables from arguments
+    """
+    envs = {}
+    # default env variables passed
+    envs['default'] = {'OMP_NUM_THREADS', 'LD_LIBRARY_PATH', 'AWS_ACCESS_KEY_ID',
+                          'AWS_SECRET_ACCESS_KEY', 'DMLC_INTERFACE'}
```
anirudh2290: Why is KMP_AFFINITY removed from this list ?

tracker/dmlc_tracker/ssh.py
```diff
@@ -20,23 +21,67 @@ def sync_dir(local_dir, slave_node, slave_dir):
         slave_node[1], local_dir, remote)
     subprocess.check_call([prog], shell = True)
 
-def get_env(pass_envs):
+def prepare_envs(args):
```
anirudh2290: Since the logic for prepare_envs seems to be common for ssh and local it would be nice to write a common function in the future which can be used by both modules.

tracker/dmlc_tracker/ssh.py
```diff
@@ -51,8 +96,13 @@ def submit(args):
...
             # hosts now contain the pair ip, port
             hosts.append((h, p))
+    return hosts
+
+def submit(args):
+    hosts = prepare_hosts(args.host_file)
+    envs = prepare_envs(args)
 
-    def ssh_submit(nworker, nserver, pass_envs):
+    def ssh_submit(nworker, nserver, dmlc_envs, addnl_envs):
```
anirudh2290: This feature can also be added to other submission scripts from local and ssh correct ?

tracker/dmlc_tracker/ssh.py
rahul003: All the common work is handled by parse_env_pairs method in opts.py 
Other than that, we have some differences between local and ssh, because in local we copy all environment variables which are set. But for ssh we pass specific ones. There's no redundant code as such here.

tracker/dmlc_tracker/ssh.py
rahul003: good catch. I added it back. I suspect it has to do with the fact that I was working on this before KMP_AFFINITY was added. Either way, brought it back.

tracker/dmlc_tracker/opts.py
rahul003: removed

tracker/dmlc_tracker/local.py
rahul003: added

tracker/dmlc_tracker/ssh.py
rahul003: I see that currently only ssh mode passes additional env variables (like OMP_NUM_THREADS, KMP_AFFINITY..). 
MPI, Yarn and SGE are only passing dmlc environment variables, which are required to start training. Which leads me to believe either setting such environment variables is not useful/important in those modes (OR) these modes aren't being used much. 
Besides I don't have access to such clusters to verify that it works. I can clarify this in mxnet/tools/launch.py .

rahul003:
Thanks @anirudh2290 for your review. Addressed them

rahul003:
@piiswrong @tqchen Can you please review or/and merge this? This is a useful feature. Some examples of when you would want to use this. Either way, this doesn't break any existing functionality and is an opt-in feature.

- If we want to profile kvstore server, we can use an environment variable to profile the server. Else it requires lot of hacking of code to do that.
- If we want to set particular OMP_NUM_THREADS for server and different for worker. This is a possibility because for small models, using many threads for gradient compression quantizing/addition is slower than just using one thread
- If we want to pass PYTHONPATH to each node without having to install mxnet systemwide. This is IMHO the most useful part of this feature.
- Setting MXNET_CUDNN_AUTOTUNE_DEFAULT for each node helps fp16 performance during distributed training.

rahul003:
@tqchen @piiswrong Thoughts please?

tracker/dmlc_tracker/ssh.py
```diff
@@ -20,23 +21,68 @@ def sync_dir(local_dir, slave_node, slave_dir):
...
     subprocess.check_call([prog], shell = True)
 
-def get_env(pass_envs):
+def prepare_envs(args):
+    """
+    Load environment variables from arguments
+    """
+    envs = {}
+    # default env variables which are always passed by the system
+    envs['default'] = {'OMP_NUM_THREADS', 'KMP_AFFINITY', 'LD_LIBRARY_PATH', 'AWS_ACCESS_KEY_ID',
```
piiswrong: I'm not sure if it makes sense to have these defaults

rahul003: These were passed earlier, so I kept them as is. If the variable is not set locally(on the machine which launch.py is running), then it has no effect anyway. LD_LIBRARY_PATH is one that seems a bit strange considering different machines might not have similar installations. But for most cases where uses are running on similar setups, it should be fine.

rahul003:
@piiswrong  I agree with what you said. The reason I made it that way was that python argparse default style for list of args is that. But since it's verbose I changed them to accept a comma separated string. 

Here's an example. 

```PYTHONPATH=../../python python ../../tools/launch.py -n 2 -H hosts --env PYTHONPATH --env-server OMP_NUM_THREADS:1 --env-worker MXNET_PROFILER_AUTOSTART:1,MXNET_ENGINE_TYPE:ThreadedEngine python train_imagenet.py --benchmark 1 --kv-store dist_sync```

As you can see below it is useful for debugging as well as more control over the environment variables for server and worker, which are both very different processes. Without ability to pass PYTHONPATH, each time I want to run distributed training with a local build, I need to go change a dmlc file to get it to work. That is a painful process, which we would rather not want the developers and users to go through.

File: tracker/dmlc_tracker/ssh.py (hunk 6 above)
Comment (rahul003) [reply tracker/dmlc_tracker/ssh.py]: Which ones do you think we should remove?

**Comment by rahul003** [comment]
CI passes
Could you review this useful feature?

**Comment by eric-haibin-lin** [comment]
can you resolve conflicts?

**Comment by Vikas89** [comment]
@rahul003 I think this has potential to break existing functionality. 
If user is using --env env1 --env env2 in his job or script , their script would start failing.

Can we please make sure that changes are backward compatible ?

**Comment by hcho3** [comment]
@Vikas89 I'm going to go ahead and revert this for now. Let's try to find a way to fix this.

**Comment by hcho3** [comment]
I submitted #466 to revert.

**Comment by rahul003** [comment]
So, I actually originally had this style of having the user specify --env for each environment variable (and --env-worker and --env-server for each variable). In my reviews, I was given feedback that it is too verbose. That's when I changed to taking a comma separated list. Also, MXNet has not documented the usage of the --env flag.

**Comment by Vikas89** [comment]
I agree that it looks like too verbose but so that we don't break any existing functionality, let's support this.
One suggestion is to take other optional parameter env-str-list for the new method of specifying env variables.

File: tracker/dmlc_tracker/ssh.py [hunk 7]
```diff
@@ -19,25 +20,71 @@ def sync_dir(local_dir, slave_node, slave_dir):
...
-    keys = ['OMP_NUM_THREADS', 'KMP_AFFINITY', 'LD_LIBRARY_PATH', 'AWS_ACCESS_KEY_ID',
-            'AWS_SECRET_ACCESS_KEY', 'DMLC_INTERFACE']
-    for k in keys:
+    for k in addnl_envs['default']:
+        v = os.getenv(k)
+        if v is not None:
+            append_env_var(envs, k, v)
+
+    for k in addnl_envs['user']:
         v = os.getenv(k)
```
Vikas89: if user has already supplied the key value in arguments, does it make sense to check os environment variables here ?
I think this can also break, please verify what happens if 
- I give --env VIKAS:1 
And Vikas is not in my environment variable, will it throw exception - "The environment variable '+ k + ' was passed but not set"

tracker/dmlc_tracker/opts.py
```diff
@@ -122,8 +144,13 @@ def get_opts(args=None):
                               ' but corresponds to archieve files that will be unziped locally,' +
                               ' You can use this option to ship python libraries.' +
                               ' Only valid in yarn jobs.'))
-    parser.add_argument('--env', action='append', default=[],
-                        help='Client and ApplicationMaster environment variables.')
```
hcho3: @rahul003 @Vikas89 I took a look at the code today and currently only `yarn.py` refers to `args.env`. The trouble is that `env` is assumed to be a list in the following line:
 https://github.com/dmlc/dmlc-core/blob/93d0122b0e864ec64b8255e56bb23b86d465460c/tracker/dmlc_tracker/yarn.py#L105-L106

So we actually do want to keep `action=append` for the `--env` argument.

tracker/dmlc_tracker/local.py
```diff
@@ -8,15 +8,43 @@
...
+def prepare_envs(args):
+    """
+    Load environment variables from arguments
+    """
+    envs = {}
+    # envs['default'] refers to the default env variables passed for both worker and server
+
+    # if --env option was used to pass a variable it would automatically be
+    # included because of the next line
+    envs['default'] = os.environ.copy()
```
hcho3: If `--env` is specified in the local tracker, we may want to add new environment variables. So
```
VAR1=VAL1 VAR2=VAL2 ./tracker/dmlc_tracker/dmlc-submit --cluster=local --env VAR3:VAL3
```
should assign `{'VAR1': 'VAL1', 'VAR2': 'VAL2', 'VAR3': 'VAL3'}` to `envs['default']`.

hcho3: Right now, it appears that `--env` is ignored in the local tracker. Let's make a decision whether to honor it in other trackers.

Some considerations:
* Originally, only the YARN tracker used to honor `--env` argument; other trackers used to ignore it.
* This pull request changes the SSH tracker to recognize `--env`.

It actually may be easier to ignore `--env` in all trackers other than the YARN tracker. This retains the original semantics, and for the local tracker user can always manually set env vars.

hcho3: My proposal: 1) Reserve `--env` for the YARN tracker, 2) introduce a new argument `--env-both` to accommodate the needs of the SSH tracker.

tracker/dmlc_tracker/opts.py
```diff
@@ -122,8 +144,13 @@ def get_opts(args=None):
...
+    parser.add_argument('--env', type=str, default='',
+                        help='Comma separated list of Client and ApplicationMaster environment \
+                        variables.')
+    parser.add_argument('--env-server', type=str, default='',
+                        help='Comma separated list of environment variable:value for server only')
+    parser.add_argument('--env-worker', type=str, default='',
+                        help='Comma separated list of environment variable:value for worker only')
```
hcho3: Let's document that `--env-server` and `--env-worker` are only valid for the SSH tracker.

tracker/dmlc_tracker/opts.py
hcho3: Also, the current docstring says `Comma separated list of Client and ApplicationMaster environment variables`, which clearly indicates that this argument is reserved for the YARN tracker.

PR #351, opened 2017-12-21
Removed specializations for float/double parsing

### LINKED ISSUE #323: FieldEntry<float> on g++5.4
While debugging dmlc/xgboost#2417 I came across a peculiar bug in `std:stof` which, when compiled on Linux with g++5.4, return 0 for `"0.5"`. Interestingly, a simple example outside of XGBoost codebase does not trigger the bug, so I suspect it is caused by some specific combination of flags... wdyt?

```diff
diff --git a/include/dmlc/parameter.h b/include/dmlc/parameter.h
index 73c2b9e..9477bf3 100644
--- a/include/dmlc/parameter.h
+++ b/include/dmlc/parameter.h
@@ -958,6 +958,7 @@ class FieldEntry<float> : public FieldEntryNumeric<FieldEntry<float>, float> {
   typedef FieldEntryNumeric<FieldEntry<float>, float> Parent;
   // override set
   virtual void Set(void *head, const std::string &value) const {
+   std::cout << (float) std::stof("0.5") << std::endl;
     try {
       this->Get(head) = std::stof(value);
     } catch (const std::invalid_argument &) {
```
```
$ g++ --version
g++ (Ubuntu 5.4.0-6ubuntu1~16.04.5) 5.4.0 20160609
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

superbobry:
These use locale-dependent API which makes little sense for parameters.

Closes #323

### DIFF
--- a/include/dmlc/parameter.h
+++ b/include/dmlc/parameter.h
@@ -975,59 +975,6 @@ class FieldEntry<bool>
     os << static_cast<int>(value);
   }
 };
-
-
-// specialize define for float. Uses stof for platform independent handling of
-// INF, -INF, NAN, etc.
-#if DMLC_USE_CXX11
-template <>
-class FieldEntry<float> : public FieldEntryNumeric<FieldEntry<float>, float> {
- public:
-  // parent
-  typedef FieldEntryNumeric<FieldEntry<float>, float> Parent;
-  // override set
-  virtual void Set(void *head, const std::string &value) const {
-    try {
-      this->Get(head) = std::stof(value);
-    } catch (const std::invalid_argument &) {
-      std::ostringstream os;
-      os << "Invalid Parameter format for " << key_ << " expect " << type_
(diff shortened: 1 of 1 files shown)


hcho3:
On one hand, this PR removed the locale dependency of parameter parsing (0,8 vs 0.8). On the other hand, this PR removes out-of-range check on float values. `dmlc::strtof()` silent clips too large / too small values, whereas `std::stof()` throws `std::out_of_range`. The problem of small parameters prompted #315. Let me think of a way to achieve both locale independence and value safety.

hcho3:
@superbobry Closing for now. I will open a new pull request to enable locale-independent parsing. One important difference will be that `std::out_of_range` will be thrown when an out-of-range value is given.

PR #352, opened 2018-01-01
Make choice of archiver configurable

nehaljwani:
At present, the build system assumes that the Unix archiv utility 'ar'
will always be present in PATH, but this might not be true if a separate
toolchain is being used. This patch makes the build system a bit more
cross-compile capable.


nehaljwani:
CI Failure is unrelated to this PR.

PR #354, opened 2018-01-09
Add more environment variables to support 3rd-party S3 implementations

solin319:
Add more environment variables to support 3rd-party S3 implementations。
S3_ENDPOINT defines the endpoint.
S3_VERIFY_SSL defines whether to close SSL certificate.

We also have changed all the parameters' name with "aws" to "s3".

Does DMLC-CORE need this feature?
@tqchen

### DIFF
--- a/src/io/s3_filesys.cc
+++ b/src/io/s3_filesys.cc
@@ -83,7 +83,7 @@ static std::string Base64(unsigned char md[], unsigned len) {
   return ret;
 }
 /*!
- * \brief sign given AWS secret key
+ * \brief sign given S3 secret key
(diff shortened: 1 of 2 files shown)


tqchen:
pls keep the env variable backward compatible, i.e. allow use of both AWS_XXX and S3_XX to set the env, where S3_XX overrides AWS_XXX

There are some lint error to be fixed in https://travis-ci.org/dmlc/dmlc-core/jobs/326766783

solin319:
Lint errors have been fixed now.
AWS_XXX can be used when S3_XX is NULL.
@tqchen

PR #357, opened 2018-01-15
Fix Clang warning on ATOMIC_FLAG_INIT braced initialization

larroy:


larroy:
@tqchen @piiswrong can we merge this? It's blocking our ARM builds. Thank you.

KellenSunderland:
Without applying this patch I'm getting the following error when building with clang:

```
In file included from src/engine/threaded_engine_pooled.cc:28:
/work/mxnet/dmlc-core/include/dmlc/concurrency.h:31:22: warning: braces around scalar initializer [-Wbraced-scalar-init]
  Spinlock() : lock_(ATOMIC_FLAG_INIT) {
                     ^~~~~~~~~~~~~~~~
/usr/aarch64-linux-android/bin/../lib/gcc/aarch64-linux-android/4.9.x/../../../../include/c++/4.9.x/atomic:1821:26: note: expanded from macro 'ATOMIC_FLAG_INIT'
#define ATOMIC_FLAG_INIT {false}
                         ^~~~~~~
```

larroy:
@cjolivier01 ideas are welcome if you have a better one. I tried to remove the braces in the constructor argument and it was commented against that. (Meaning removing the parenthesis on the constructor initialization list so the brakets are not surrounded by parens).  I don't this we should spend too much time discussing about that as disabling the warning explictily is a valid approach IMHO.

cjolivier01:
For example (there's already a test there for MSVC):

class Spinlock {
 public:
#if defined(_MSC_VER) || defined(__clang__)
  Spinlock() {
    lock_.clear();
  }
#else
  Spinlock() : lock_(ATOMIC_FLAG_INIT) {
  }
#endif

larroy:
I think that's a worse solution, to duplicate code for a warning than what I propose.

cjolivier01:
what do you mean by “duplicate”? that code is already duplicated for MSVC compiler.

Issue #358, opened 2018-01-17
LocalFileSystem::ListDirectory() fails on directories with dangling symlinks


DickJC123:
ListDirectory expects to convert a directory to a vector<FileInfo>.  However, if the directory being listed contains a dangling symlink, the call fails with a fatal error 'no such file or directory'.  Normally, one would expect this to be uncommon, however emacs creates dangling symlinks as part of its file locking strategy.  Thus, a user emacs-editing a file (with a dirty buffer) would disrupt the proper operation of ListDirectory if run on the same directory.

PR #359, opened 2018-01-18
ListDirectory to warn rather than fatal-exit on dangling symlinks

DickJC123:
A fix for issue #358.

ListDirectory expects to convert a directory to a vector of FileInfo. However, if the directory being listed contains a dangling symlink, the call fails with a fatal error 'no such file or directory'. Normally, one would expect this to be uncommon, however emacs creates dangling symlinks as part of its file locking strategy. Thus, a user emacs-editing a file (with a dirty buffer) would disrupt the proper operation of ListDirectory if run on the same directory.

The symlink issue is encountered in LocalFileSystem::GetPathInfo(), but that routine must return a FileInfo.  This fix thus has GetPathInfo map the fatal dangling symlink to a FileInfo depicting a 0-size file- such files are ignored by the current uses of ListDirectory().

### DIFF
--- a/src/io/local_filesys.cc
+++ b/src/io/local_filesys.cc
@@ -69,13 +69,22 @@ class FileStream : public SeekStream {
 
 FileInfo LocalFileSystem::GetPathInfo(const URI &path) {
   struct stat sb;
-  if (stat(path.name.c_str(), &sb) == -1) {
-    int errsv = errno;
-    LOG(FATAL) << "LocalFileSystem.GetPathInfo " << path.name
-               << " Error:" << strerror(errsv);
-  }
   FileInfo ret;
   ret.path = path;
+  if (stat(path.name.c_str(), &sb) == -1) {
+    int errsv = errno;
+    // If lstat succeeds where stat failed, assume a problematic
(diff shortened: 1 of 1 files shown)


src/io/local_filesys.cc
```diff
@@ -69,13 +69,22 @@ class FileStream : public SeekStream {
...
   struct stat sb;
+  FileInfo ret;
+  ret.path = path;
   if (stat(path.name.c_str(), &sb) == -1) {
     int errsv = errno;
-    LOG(FATAL) << "LocalFileSystem.GetPathInfo " << path.name
-               << " Error:" << strerror(errsv);
+    // If lstat succeeds where stat failed, assume a problematic
+    // symlink and treat this as if it were a 0-length file.
+    if (lstat(path.name.c_str(), &sb) == 0) {
```
anirudh2290: @DickJC123 This breaks MXNet CI tests on Windows. Please see: http://jenkins.mxnet-ci.amazon-ml.com/blue/organizations/jenkins/incubator-mxnet/detail/PR-9869/6/pipeline

tqchen:
this change is reverted for now for windows compatibility issue

PR #366, opened 2018-02-14
Better Exception Handling for Iterators

anirudh2290:
# Description

* Adds exception handling support for iterators.
* Exceptions thrown in beforefirst, next of the producer in threadediter will be caught and rethrown.
* Once the exception is set, it will always be rethrown for all following calls of next and beforefirst of the consumer. Exception_ptr will be reset to nullptr on another call of Init. 
* In addition to this, I have also added exception handling inside omp threads in TextParser.
* This should improve the usability experience of MXNet, which heavily uses ThreadedIter and TextParser.

# Testing

* Tested for scenarios where exception is thrown from producer's beforefirst and next and rethrown from next, beforefirst, recycle.
* Tested consecutive calls of beforefirst to make sure exception_ptr not cleared.
* Tested Init call after exception to test if its cleared.
* Will add additional tests to MXNet for TextParser.

@piiswrong @KellenSunderland @reminisce @madjam

### DIFF
--- a/include/dmlc/threadediter.h
+++ b/include/dmlc/threadediter.h
@@ -145,6 +145,11 @@ class ThreadedIter : public DataIter<DType> {
   inline void ThrowExcIfSet(void);
 
   /*!
+   * \brief clears exception_ptr, called from Init
+   */
+  inline void EmptyExc(void);
+
+  /*!
    * \brief adapt the iterator interface's Next
    *  NOTE: the call to this function is not threadsafe
    *  use the other Next instead
@@ -297,6 +302,7 @@ inline void ThreadedIter<DType>::Init(std::function<bool(DType **)> next,
   producer_sig_ = kProduce;
   producer_sig_processed_ = false;
   produce_end_ = false;
+  EmptyExc();
   // procedure running in prodcuer
   // run producer thread
   auto producer_fun = [this, next, beforefirst]() {
@@ -374,6 +380,7 @@ inline void ThreadedIter<DType>::Init(std::function<bool(DType **)> next,
             iter_exception_ = std::current_exception();
           }
         }
+        bool next_notify = false;
         {
           std::unique_lock<std::mutex> lock(mutex_);
           if (producer_sig_ == kBeforeFirst) {
@@ -381,14 +388,19 @@ inline void ThreadedIter<DType>::Init(std::function<bool(DType **)> next,
               free_cells_.push(queue_.front());
               queue_.pop();
             }
-            produce_end_ = false;
+            produce_end_ = true;
             producer_sig_processed_ = true;
-            producer_sig_ = kProduce;
             lock.unlock();
             consumer_cond_.notify_all();
-            return;
+          } else if (producer_sig_ == kProduce) {
+            produce_end_ = true;
+            next_notify = nwait_consumer_ != 0;
+            lock.unlock();
+            if (next_notify)
+              consumer_cond_.notify_all();
           }
         }
+        return;
       }
     }
   };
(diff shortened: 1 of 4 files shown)


piiswrong:
why do you need a mutex?

anirudh2290:
@piiswrong To prevent the race condition caused by iter_exception_ being written by the producer the same time consumer calls the Next and checks if iter_exception_ is set.

anirudh2290:
@piiswrong Any comments ?

anirudh2290:
I realized that there is one drawback here. If you have MXNet code with something like: 
```
try:
     data_train = mx.io.CSVIter(data_csv="data.csv", label_csv="label.csv", data_shape=(1, 10),
                                                 batch_size=4)
except mx.base.MXNetError as ex:
     print("Exception handled")
     print(ex)
```
Below is data.csv:
```
1,2,3,4,5,6,7,8,9,10
1,2,3,4,5,6,7,8,9,10
1,2,3,4,5,6,7,8,9,10
1,2,3,4,5,6,7,8,9,10
1,2,3,4,5,6,7,8,9,10
1,2,3,4,5,6,7,8,9,10
1,2,3,4,5,6,7,8,9,10
1,2,3,4,5,6,7,8,9,10
```

and below is label.csv:

```
label1
label2
label3
label4
label5
label6
label7
```

There will be an exception thrown when prefetching the second batch, since the number of labels are less than the number of samples in data.csv. This will set the exception_ptr iter_exception_ to be thrown by next(consumer). 

The problem is that in the above case there is only one call to next (consumer) which happens as part of the mx.io.CSVIter instantiation. 

There is a race condition here, because depending on when the next(consumer) is executed, before the iter_exception_ is set or after the iter_exception_ is set we will end up throwing or not throwing the exception.

This non-deterministic behavior will only happen when user is making next calls which are less in number compared to index of the batch where exception is thrown. 

For most use cases we iterate through all of the dataset, this should not be a problem. 

@piiswrong Please advise on whether its okay to not handle this usecase.

include/dmlc/threadediter.h
```diff
@@ -138,6 +138,17 @@ class ThreadedIter : public DataIter<DType> {
    *        the content of inout_dptr will be set to NULL
    */
   inline void Recycle(DType **inout_dptr);
+
+  /*!
+   * \brief Rethrows exception which is set by the producer
+   */
+  inline void ThrowExcIfSet(void);
```
piiswrong: exc->exception

include/dmlc/threadediter.h
```diff
@@ -138,6 +138,17 @@ class ThreadedIter : public DataIter<DType> {
...
+
+  /*!
+   * \brief clears exception_ptr, called from Init
+   */
+  inline void EmptyExc(void);
```
piiswrong: ClearException

include/dmlc/threadediter.h
```diff
@@ -278,120 +295,181 @@ Init(Producer *producer, bool pass_ownership) {
...
   producer_thread_ = new std::thread(producer_fun);
 }
 
-template<typename DType>
-inline bool ThreadedIter<DType>::
-Next(DType **out_dptr) {
-  if (producer_sig_ == kDestroy) return false;
+template <typename DType>
+inline bool ThreadedIter<DType>::Next(DType **out_dptr) {
+  if (producer_sig_ == kDestroy)
```
piiswrong: why would this be called during destruction?

anirudh2290: this was not added by my code.

include/dmlc/threadediter.h
anirudh2290: will change

include/dmlc/threadediter.h
anirudh2290: will change

anirudh2290:
@piiswrong Thanks for your comments! I have made the requested changes.

anirudh2290:
@piiswrong ping

include/dmlc/threadediter.h
tqchen: consumer thread may not notice there is destruction happening. The producer thread need to signal destruction and consumer receives the signal and then exit

anirudh2290: @tqchen Thank you for the explanation! Since there is only a single consumer and the Destroy method is called by the consumer, which results in signal to producer and then notifying Consumer. I am not sure how Next is being called here.  "The producer thread need to signal destruction and consumer receives the signal and then exit". Can you please elaborate on this.  Its not waiting for any signal on this line : `if (producer_sig_ == kDestroy)`

tqchen: OK, I think that was a behavior which is not necessary. As it is always the consumer who calls the destroy. The check is just a safeguard for undesired usecase

anirudh2290: thanks for the clarification!

PR #367, opened 2018-02-15
Handle Windows killing all non-main threads on ExitProcess() called

cjolivier01:
Auto-remove threads are problematic on Windows during shutdown because Windows kills all threads immediately when ExsitProcess() is called, so the threads can;t auto-remove themselves.
Solution is to allow the threads to be set to non-auto-remove so that they can be joined during cleanup (they're already dead, so the join is trivial).

### DIFF
--- a/include/dmlc/thread_group.h
+++ b/include/dmlc/thread_group.h
@@ -206,6 +206,15 @@ class ThreadGroup {
     }
 
     /*!
+     * \brief Make the thread joinable (by removing the auto_remove flag)
+     * \warning Care should be taken not to cause a race condition between this call
+     *          and parallel execution of this thread auto-removing itself
+     */
+    void make_joinable() {
+      auto_remove_ = false;
+    }
+
+    /*!
      * \brief Check whether the thread is joinable
      * \return true if the thread is joinable
      */
@@ -291,7 +300,7 @@ class ThreadGroup {
      * \brief Whether to automatically remove this thread's object from the ThreadGroup when the
      *        thread exists (perform its own cleanup)
      */
-    bool auto_remove_;
+    volatile bool auto_remove_;
   };
 
   /*!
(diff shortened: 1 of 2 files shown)


include/dmlc/thread_group.h
```diff
@@ -395,38 +404,46 @@ class ThreadGroup {
    */
   inline void join_all() {
     CHECK_EQ(!is_this_thread_in(), true);
-    std::unordered_set<std::shared_ptr<Thread>> working_set;
-    {
-      ReadLock guard(m_);
-      for (auto iter = threads_.begin(), e_iter = threads_.end(); iter != e_iter; ++iter) {
-        if (!(*iter)->is_auto_remove()) {
-          working_set.emplace(*iter);
+    do {
```
cjolivier01: Here is just a scoped addition of td::unique_lock<std::mutex> lk(join_all_mtx_);

include/dmlc/thread_group.h
```diff
@@ -291,7 +300,7 @@ class ThreadGroup {
      * \brief Whether to automatically remove this thread's object from the ThreadGroup when the
      *        thread exists (perform its own cleanup)
      */
-    bool auto_remove_;
+    volatile bool auto_remove_;
```
cjolivier01: Now it can change

include/dmlc/thread_group.h
```diff
@@ -205,6 +205,15 @@ class ThreadGroup {
       return auto_remove_;
     }
 
+    /*!
+     * \brief Make the thread joinable (by removing the auto_remove flag)
+     * \warning Care should be taken not to cause a race condition between this call
+     *          and parallel execution of this thread auto-removing itself
+     */
+    void make_joinable() {
+      auto_remove_ = false;
```
cjolivier01: not auto-remove anymore

tqchen:
It appears that the new unit-testcase for thread group will indeterministic fail for cases, see checks for history https://github.com/dmlc/dmlc-core/pull/368

@cjolivier01 can you look into to fix the issue?

cjolivier01:
ok

cjolivier01:
which platform? do you have a link to a failed build?

tqchen:
https://travis-ci.org/dmlc/dmlc-core/jobs/344620203

cjolivier01:
thank you. this isn’t actually due to the latest changes, it’s something that checks that the timer doesn’t occur too many times within a given period of time as well as making sure it doesn’t happen far too little in the time period. i think i expect 50 but check for at least 20, but i suppose the build machine under heavy load doesn’t timeslice it enough.
i will reduce the threshold to 10 from 20.

Issue #364, opened 2018-02-08
MXNet MultiNode: when using ssh launcher, ps-lite will use wrong ip even if we specify correct ip in hosts


threeleafzerg:
Hi,
I encountered the issue when train MXNet multinode.
This is my host file:
192.168.20.84
192.168.20.85
192.168.20.86
192.168.20.87
192.168.20.88
This is my running command:
python tools/launch.py -n 4 -s 1 --launcher ssh -H hosts python example/image-classification/train_tinyimagenet.py --kv-store dist_sync
However, my ps-lite log shows:
[09:28:57] src/van.cc:56: assign rank=9 to node role=worker, ip=192.168.10.88, port=37380, is_recovery=0
[09:28:57] src/van.cc:56: assign rank=11 to node role=worker, ip=192.168.10.87, port=38447, is_recovery=0
[09:28:57] src/van.cc:56: assign rank=13 to node role=worker, ip=192.168.10.86, port=59958, is_recovery=0
[09:28:57] src/van.cc:56: assign rank=15 to node role=worker, ip=192.168.10.85, port=52332, is_recovery=0
[09:28:57] src/van.cc:56: assign rank=8 to node role=server, ip=192.168.10.84, port=38232, is_recovery=0
After investigation, I found that since ssh.py didn't pass env DMLC_NODE_HOST to ps-lite. ps-lite will try available interface (eth0)and available ip(192.168.10.85). But that's not my expected interface(eth1) and available ip (192.168.20.85) I suggest ssh.py pass the env DMLC_NODE_HOST to ps-lite. 

Following is my suggested fix. (bold font) Can anyone help that.

--- a/tracker/dmlc_tracker/ssh.py
+++ b/tracker/dmlc_tracker/ssh.py
@@ -72,6 +72,7 @@ def submit(args):
         for i in range(nworker + nserver):
             pass_envs['DMLC_ROLE'] = 'server' if i < nserver else 'worker'
             (node, port) = hosts[i % len(hosts)]
             **pass_envs['DMLC_NODE_HOST'] = node**
             prog = get_env(pass_envs) + ' cd ' + working_dir + '; ' + (' '.join(args.command))
             prog = 'ssh -o StrictHostKeyChecking=no ' + node + ' -p ' + port + ' \'' + prog + '\''
             thread = Thread(target = run, args=(prog,))

threeleafzerg:
@pengzhao-intel

pengzhao-intel:
[...]

PR #370, opened 2018-02-23
Pass DMLC_NODE_HOST env to worker or server in ssh launcher

threeleafzerg:
To solve issue #364:
We hope pass correct DMLC_NODE_HOST to ps-lite. In default, this env is *not* passed and the ip of default interface is used.


pengzhao-intel:
@threeleafzerg could you help merge?

PR #373, opened 2018-02-27
Add nostream json parser

nhynes:
Requires upstreaming changes to resolve https://github.com/intel/linux-sgx/issues/218.

This PR adds a no-stream implementation of the JSON parser to allow use of the TVM graph runtime in SGX.

The code passes the tests used for the original parser. One question is whether the quite-similar files should be merged using conditional compilation at the expense of clarity.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,8 +14,8 @@ NOLINT_FILES = --exclude_path include/dmlc/concurrentqueue.h include/dmlc/blocki
 # this is the common build script for dmlc lib
 export LDFLAGS= -pthread -lm
 export CFLAGS = -O3 -Wall -Wno-unknown-pragmas -Iinclude  -std=c++0x
-LDFLAGS+= $(DMLC_LDFLAGS)
-CFLAGS+= $(DMLC_CFLAGS)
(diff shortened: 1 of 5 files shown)


tqchen:
I think a merged code is preferred, we can narrow down the code to a few basic function member, like getline, and redirects all the code to one place

tqchen:
@nhynes can you try to make a merged version? ≈

nhynes:
@tqchen Yes, absolutely! I'm still waiting on https://github.com/intel/linux-sgx/issues/218, though, to determine whether the `unordered_map` needs to be replaced with something less efficient.

My current approach, however, is to make a conditionally compiled functional mock of the required iostreams (not patching `std`, of course) so to maintain the current `JSON(Read|Writ)er` interface.

My goal is to have this (and the upstream changes) completed by no later than Thursday.

nhynes:
Okay, this should be ready to go.

To run the SGX tests (which  do pass), just set
`ADD_CFLAGS = -D_LIBCPP_SGX_NO_IOSTREAMS=1 -D_LIBCPP_SGX_CONFIG=1`
in `make/config.mk`.

PR #376, opened 2018-03-01
Android 64 fix openmp weak symbols throw specification

larroy:


include/dmlc/omp.h
```diff
@@ -5,15 +5,23 @@
...
+
+
 #if defined(_OPENMP)
 #include <omp.h>
 #else
+
+#if defined(__ANDROID__) && defined(__AARCH64EL__)
+#define __GOMP_NOTHROW
+#elif defined(__cplusplus)
+#define __GOMP_NOTHROW throw()
```
lebeg: Maybe noexcept instead?

larroy: This was a fix by somebody else, not sure if it can cause problems if we change to noexcept. I would leave it as it is.

larroy:
@tqchen

PR #378, opened 2018-03-07
AWS Signature version 4 (SIG4) Support for S3 Read and Write

rahul003:
Updated S3 File system to support SIG4. Replaces SIG2 because that is only supported in us-east-1 currently and SIG4 is supported in all regions.

Updated the original PR to fix both reads and writes

### DIFF
--- a/src/io/s3_filesys.cc
+++ b/src/io/s3_filesys.cc
@@ -67,112 +67,14 @@ struct XMLIter {
   }
 };
 
-/*!
- * \brief return a base64 encoded string
- * \param md the data
- * \param len the length of data
- * \return the encoded string
- */
-static std::string Base64(unsigned char md[], unsigned len) {
-  // encode base64
-  BIO *fp = BIO_push(BIO_new(BIO_f_base64()),
-                       BIO_new(BIO_s_mem()));
-  BIO_write(fp, md, len);
-  BIO_ctrl(fp, BIO_CTRL_FLUSH, 0, NULL);
-  BUF_MEM *res;
-  BIO_get_mem_ptr(fp, &res);
-  std::string ret(res->data, res->length - 1);
-  BIO_free_all(fp);
-  return ret;
-}
-
-/*!
- * \brief sign given S3 secret key
- * \param secret_key the key to compute the sign
- * \param content the content to sign
- */
-static std::string Sign(const std::string &key, const std::string &content) {
-  HMAC_CTX ctx;
-  unsigned char md[EVP_MAX_MD_SIZE];
-  unsigned rlen = 0;
-  HMAC_CTX_init(&ctx);
-  HMAC_Init(&ctx, key.c_str(), key.length(), EVP_sha1());
-  HMAC_Update(&ctx,
-              reinterpret_cast<const unsigned char*>(content.c_str()),
-                content.length());
-  HMAC_Final(&ctx, md, &rlen);
-  HMAC_CTX_cleanup(&ctx);
-  return Base64(md, rlen);
-}
-
-/*!
- * \brief Returns signature according to AWS Signature Version 2
- * \param key s3_access_key
- * \param method GET/PUT/POST
- * \param content_md5 hash of content
- * \param content_type type of content
- * \param date date
- * \param amz_headers headers
- * \param resource address of resource
- * \return signature as a string which is base64 encoded
- */
-static std::string Sign(const std::string &key,
-                        const std::string &method,
-                        const std::string &content_md5,
-                        const std::string &content_type,
-                        const std::string &date,
-                        std::vector<std::string> amz_headers,
-                        const std::string &resource) {
-  std::ostringstream stream;
-  stream << method << "\n";
-  stream << content_md5 << "\n";
-  stream << content_type << "\n";
-  stream << date << "\n";
-  std::sort(amz_headers.begin(), amz_headers.end());
-  for (size_t i = 0; i < amz_headers.size(); ++i) {
-    stream << amz_headers[i] << "\n";
+static const std::string SHA256HashToHex(unsigned char *hash, int size) {
+  CHECK_EQ(size, SHA256_DIGEST_LENGTH);
+  std::stringstream ss;
+  for(int i = 0; i < SHA256_DIGEST_LENGTH; i++)
+  {
+    ss << std::hex << std::setw(2) << std::setfill('0') << (int)hash[i];
   }
-  stream << resource;
-  return Sign(key, stream.str());
-}
-
-/*!
- * \brief Generates hash of input as per SHA256 algorithm
- * Ref: http://stackoverflow.com/questions/2262386/generate-sha256-with-openssl-and-c
- * \param str input to hash
- * \param outputBuffer hashed output
- */
-static void SHA256(const std::string &str, unsigned char *outputBuffer) noexcept {
-  char *c_string = new char[str.length() + 1];
-  std::snprintf(c_string, str.length() + 1, "%s", str.c_str());
-  unsigned char hash[SHA256_DIGEST_LENGTH];
-  SHA256_CTX sha256;
-  SHA256_Init(&sha256);
-  SHA256_Update(&sha256, c_string, strlen(c_string));
-  SHA256_Final(hash, &sha256);
-  for (int i = 0; i < SHA256_DIGEST_LENGTH; i++) {
-    outputBuffer[i] = hash[i];
-  }
-}
-
-/*!
- * Converts a sha256 hash char array to hex
- *
- * \param input unsinged char array
- * \param size size of input buffer
- * \return string in hex representation
- */
-static const std::string ConvertSHA256HashToHex(const unsigned char *input, const int size) {
-  CHECK_EQ(size, SHA256_DIGEST_LENGTH) << "SHA256Hash needs to be of size "
-                                       << SHA256_DIGEST_LENGTH;
-  // size for output buffer which has hex data
-  char outputBuffer[65];
-  for (int i = 0; i < size; i++) {
-    // converting to hex
-    snprintf(outputBuffer + (i * 2), sizeof(outputBuffer), "%02x", input[i]);
-  }
-  outputBuffer[64] = 0;
-  return std::string{outputBuffer};
+  return ss.str();
 }
 
 /*!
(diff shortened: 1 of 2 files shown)


rahul003:
@cjolivier01 @anirudh2290 Could you help review?

src/io/s3_filesys.cc
```diff
@@ -121,6 +136,235 @@ static std::string Sign(const std::string &key,
   return Sign(key, stream.str());
 }
 
+/*!
+ * \brief Generates hash of input as per SHA256 algorithm
+ * Ref: http://stackoverflow.com/questions/2262386/generate-sha256-with-openssl-and-c
+ * \param str input to hash
+ * \param outputBuffer hashed output
+ */
+static void SHA256(const std::string &str, unsigned char *outputBuffer) noexcept {
```
cjolivier01: how do you assure outputBuffer is the correct size or large enough?

src/io/s3_filesys.cc
```diff
@@ -121,6 +136,235 @@ static std::string Sign(const std::string &key,
...
+                                           const std::string &secret,
+                                           const std::string &region,
+                                           const std::string &service,
+                                           const std::string &string_to_sign) noexcept {
+  const std::string k1{"AWS4" + secret};
+  char *c_k1 = new char[k1.length() + 1];
+  std::snprintf(c_k1, k1.length() + 1, "%s", k1.c_str());
+
+  auto yyyymmdd = GetDateYYYYMMDD(request_date);
+  char *c_yyyymmdd = new char[yyyymmdd.length() + 1];
```
cjolivier01: why aren’t these char * used with unique_ptr? who frees them?

src/io/s3_filesys.cc
```diff
@@ -492,33 +769,47 @@ class ReadStream : public CURLReadStreamBase {
 void ReadStream::InitRequest(size_t begin_bytes,
                              CURL *ecurl,
                              curl_slist **slist) {
-  // initialize the curl request
-  std::vector<std::string> amz;
+  std::string payload = "";
+  time_t curr_time = time(NULL);
```
cjolivier01: you don’t ever need to assign to an empty string. in fact, it adds unnecessary overhead

src/io/s3_filesys.cc
rahul003: Is this a concern even though I call this from one location where I know the size required. Should I also pass size of output buffer and add a check?

src/io/s3_filesys.cc
```diff
@@ -192,6 +429,37 @@ std::string getEndpoint(std::string region_name) {
...
+      encoded_str << c;
+    } else if (c == '/') {
+      if (encodeSlash) {
+        encoded_str << "%2F";
+      } else {
+        encoded_str << c;
+      }
+    } else {
+      encoded_str << '%';
+      encoded_str << std::setw(2) << static_cast<unsigned>(c);
```
anirudh2290: Wont this just output % followed by decimal value ? Dont you want the hexadecimal representation ?

rahul003: No, because of the usage of std::hex. Which it is set as basefield, integer values inserted into the stream are expressed in hexadecimal base.

An example
```
#include <iostream>  
int main () {
  int n = 70;
  std::cout << std::dec << n << '\n';
  std::cout << std::hex << n << '\n';
  std::cout << std::oct << n << '\n';
  return 0;
}
```
Output
```
70
46
106
```

src/io/s3_filesys.cc
```diff
@@ -121,6 +136,226 @@ static std::string Sign(const std::string &key,
...
+                  reinterpret_cast<const unsigned char*>(AWS4_REQUEST.c_str()),
+                  AWS4_REQUEST.size(), NULL, &kSigningLen);
+
+  unsigned char *kSig;
+  unsigned int kSigLen;
+  kSig = HMAC(EVP_sha256(), kSigning, strlen(reinterpret_cast<char *>(kSigning)),
+              reinterpret_cast<const unsigned char*>(string_to_sign.c_str()),
+              string_to_sign.size(), NULL, &kSigLen);
+  // convert to hex
+  char outputBuffer[65];
```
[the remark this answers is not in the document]
rahul003: Done

src/io/s3_filesys.cc
```diff
@@ -121,6 +136,226 @@ static std::string Sign(const std::string &key,
...
+ * \param service AWS service name
+ * \param string_to_sign
+ * \return signature
+ */
+static const std::string CalculateSig4Sign(const std::time_t &request_date,
+                                           const std::string &secret,
+                                           const std::string &region,
+                                           const std::string &service,
+                                           const std::string &string_to_sign) noexcept {
+  const std::string k1{"AWS4" + secret};
```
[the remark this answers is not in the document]
rahul003: key1

src/io/s3_filesys.cc
anirudh2290: Thanks for explaining. I missed that line.

rahul003:
Regarding tests, I'm not sure how to make an end-to-end test because that requires secret_key , access_id, creating some buckets, etc. We do have a s3filesystest from before, and since the interface hasn't changed, that could be a decent test for this feature.

rahul003:
I've updated this PR to support writes as well

rahul003:
Also fixes a bug when trying to load path like `s3://bucketname/file.csv`

szha:
@DamonDeng

DamonDeng:
Got it, I will review the code and test it.

DamonDeng:
Tried to test the pull request, triggered an issue caused by OpenSSL.

While calling BIO_get_mem_ptr(fp, &res) in Base64 function, an BUF_MEM object with huge length will be returned, and then causes memory allocation fatal. 

	static std::string Base64(unsigned char md[], unsigned len) {
	  // encode base64
	  BIO *fp = BIO_push(BIO_new(BIO_f_base64()),
	                       BIO_new(BIO_s_mem()));
	  BIO_write(fp, md, len);
	  BIO_ctrl(fp, BIO_CTRL_FLUSH, 0, NULL);
	  BUF_MEM *res;
	  BIO_get_mem_ptr(fp, &res);
	  
	  // Memory allocation fatal occurs here.
	  std::string ret(res->data, res->length - 1);
	  BIO_free_all(fp);
	  return ret;
	} 
	
Then I found that this issue is not related to this pull request, with or without this pull request, the memory allocation fatal will occur.

@rahul003 

Is there any special requirements of the environment to test this pull request?

It will be appreciated if you can post the python script you used to test the S3 reading and S3 writing.
You don't need to provide your access key and secret key, I will modify the script to use my own keys and S3 buckets.

rahul003:
Hi @DamonDeng ,
Are you sure you're running this code? After the change, there should be no usage of `BIO_...`.
I used this script to test it. You can modify the locations and test. 
https://github.com/rahul003/mxnet/blob/s3-auth/tests/python/train/test_conv.py

You need to set S3_REGION, S3_ACCESS_KEY_ID, and S3_SECRET_ACCESS_KEY to use it

DamonDeng:
@rahul003 , sorry to tell you that I tested the wrong branch.
I checked out the right one and tested it again with my own testing script.

The pull request works well for all the AWS Regions, including 15 global regions, and 2 China regions.

I would like to recommend the team to merge your pull request.

Well Done! @rahul003

rahul003:
Thanks @DamonDeng for verifying.

@tqchen Can you help review and merge this?

rahul003:
@tqchen @piiswrong Could you take a look and merge this?

src/io/s3_filesys.cc
```diff
@@ -64,70 +66,289 @@ struct XMLIter {
...
+  }
+  return ss.str();
+}
+
+/*!
+ * \brief Generates hash of input as per SHA256 algorithm and converts it to hex representation
+ * \param str input to hash
+ * \return string with hex representation of SHA256 Hash
+ */
+static const std::string SHA256Hex(const std::string &str) noexcept {
```
cjolivier01: const in return is superfluous

src/io/s3_filesys.cc
```diff
@@ -64,70 +66,289 @@ struct XMLIter {
...
+ * \brief Converts hash to hex representation
+ * \param hash unsigned char array with hash
+ * \param size size of hash
+ * \return string in hex representation
+ */
+static const std::string SHA256HashToHex(unsigned char *hash, int size) {
+  CHECK_EQ(size, SHA256_DIGEST_LENGTH);
+  std::stringstream ss;
+  for (int i=0; i < SHA256_DIGEST_LENGTH; i++) {
+    ss << std::hex << std::setw(2) << std::setfill('0') << static_cast<int>(hash[i]);
```
cjolivier01: should you save current formatting modes and restore?

src/io/s3_filesys.cc
```diff
@@ -64,70 +66,289 @@ struct XMLIter {
...
-  BIO *fp = BIO_push(BIO_new(BIO_f_base64()),
-                       BIO_new(BIO_s_mem()));
-  BIO_write(fp, md, len);
-  BIO_ctrl(fp, BIO_CTRL_FLUSH, 0, NULL);
-  BUF_MEM *res;
-  BIO_get_mem_ptr(fp, &res);
-  std::string ret(res->data, res->length - 1);
-  BIO_free_all(fp);
-  return ret;
+static const std::string GetDateYYYYMMDD(const std::time_t &t) noexcept {
```
cjolivier01: superfluous const

src/io/s3_filesys.cc
```diff
@@ -64,70 +66,289 @@ struct XMLIter {
...
-                content.length());
-  HMAC_Final(&ctx, md, &rlen);
-  HMAC_CTX_cleanup(&ctx);
-  return Base64(md, rlen);
+std::string URIEncode(const std::string& str,
+                      bool encodeSlash = true) {
+  std::stringstream encoded_str;
+  encoded_str << std::hex << std::uppercase << std::setfill('0');
+  for (std::string::const_iterator it = str.begin(); it != str.end(); ++it) {
+    char c = *it;
```
cjolivier01: next time, consider a switch

src/io/s3_filesys.cc
```diff
@@ -757,10 +962,12 @@ void WriteStream::Run(const std::string &method,
     LOG(FATAL) << "AWS S3 Error:\n" << *out_header << *out_data;
   }
 }
+
 void WriteStream::Init(void) {
   std::string rheader, rdata;
-  Run("POST", path_, "?uploads",
-      "binary/octel-stream", "", &rheader, &rdata);
+  std::map<std::string, std::string> params;
+  params["uploads"] = "";
```
cjolivier01: nit: you could just call
clear() but this is ok

rahul003:
@piiswrong Could you also look at this? Without this change, only the use of us-east-1 region works for S3 currently.

src/io/s3_filesys.cc
rahul003: Remove thanks

src/io/s3_filesys.cc
rahul003: I couldn't find a reference on if that is possible. Besides this is a static method...

PR #384, opened 2018-03-13
fix bug in s3 read retry

solin319:
We think the parameter used in FillBuffer should be 'nleft'. 
When we train imagenet-classification with s3file, this->FillBuffer(1) != 0 usually cause 50 times failed retry.
And we change 1 to 'nleft' to fix this bug.
Is this method right?
@tqchen


tqchen:
LGTM

PR #401, opened 2018-05-17
Add indexing_mode option to libsvm and libfm parser

hcho3:
Currently the LibSVM and LibFM parsers assume 0-based indexing. However, many LibSVM files in the wild use 1-based indexing. (In fact, the [original LIBSVM implementation](https://github.com/cjlin1/libsvm/blob/master/README) uses 1-based indexing.) This PR adds a new option `indexing_mode` to both parsers to enable 1-based indexing.

Description (EDITED):
* `indexing_mode>0`: use 1-based indexing.
* `indexing_mode=0`: use 0-based indexing
* `indexing_mode<0`: use heuristic to decide between the two modes of indexing. When the smallest feature index is 1 or greater, 1-based indexing is chosen.

For examples, see the unit tests in `test/unittest/unittest_parser.cc`.

Note. Backward compatibility is preserved as follows: when `indexing_mode` is not specified, 0-based indexing is assumed.

### DIFF
--- a/src/data.cc
+++ b/src/data.cc
@@ -26,7 +26,7 @@ CreateLibSVMParser(const std::string& path,
                    unsigned num_parts) {
   InputSplit* source = InputSplit::Create(
       path.c_str(), part_index, num_parts, "text");
-  ParserImpl<IndexType> *parser = new LibSVMParser<IndexType>(source, 2);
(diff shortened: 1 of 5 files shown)


hcho3:
Now that I think about, it may be better to use `indexing_mode>0` to indicate 1-based, `indexing_mode=0` to indicate 0-based, and `indexing_mode<0` to automatic detection.

PR #406, opened 2018-05-30
[IO] Enable endian aware serialization

tqchen:
The original binary serialization is not aware of endianess of the target machine. This patch resolves that by add endian aware serialization and loading. 

Default behavior matches the existing serialization(use little endian) on most of machines we target


tqchen:
cc @piiswrong

PR #411, opened 2018-06-18
Fix 'make lint' build issue for python2 users

apeforest:
An earlier commit (https://github.com/dmlc/dmlc-core/commit/b744643f386660ddc39467a04e3a98853a7419b9) replaced pip to pip3 in the .travis.yml. While it made pip3 work well it broke the 'make lint' command for python2 users. I made a change in .travis.yml to support multiple python versions. This practice is borrowed from cpplint package (https://github.com/cpplint/cpplint/blob/master/.travis.yml). 

Verified the fix locally by installing cpplint and pylint using pip(2).


tqchen:
@szha I would recommend only enable pylint for python3 from this point on, as python2 is going to be deprecated by major packages in 2019

apeforest:
@tqchen I am okay with that but we probably need to make this clear somewhere in the README. If you guys all agree, I could do just that too.

apeforest:
This has been fixed by https://github.com/dmlc/mshadow/pull/340. Close this PR.

PR #417, opened 2018-06-25
Add support for int64 for CSVParser

haojin2:
Same as title.
This PR adds support for parsing int64 data type in dmlc-core, so that mxnet could also support this type in CSVIter

### DIFF
--- a/src/data/csv_parser.h
+++ b/src/data/csv_parser.h
@@ -88,11 +88,14 @@ ParseBlock(const char *begin,
       if (std::is_same<DType, real_t>::value) {
         v = strtof(p, &endptr);
       // If DType is int32
-      } else if (std::is_same<DType, int>::value) {
-        v = static_cast<int>(strtol(p, &endptr, 0));
+      } else if (std::is_same<DType, int32_t>::value) {
+        v = static_cast<int32_t>(strtol(p, &endptr, 0));
+      // If DType is int64
+      } else if (std::is_same<DType, int64_t>::value) {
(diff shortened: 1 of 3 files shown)


src/data/csv_parser.h
```diff
@@ -90,9 +90,12 @@ ParseBlock(const char *begin,
       // If DType is int32
       } else if (std::is_same<DType, int>::value) {
         v = static_cast<int>(strtol(p, &endptr, 0));
+      // If DType is int64
+      } else if (std::is_same<DType, int64_t>::value) {
+        v = static_cast<int64_t>(strtol(p, &endptr, 0));
       // If DType is all other types
       } else {
-        LOG(FATAL) << "Only float32 and int32 are supported for the time being";
+        LOG(FATAL) << "Only float32, int32, and int64 are supported for the time being";
```
hcho3: Is it reasonable to assume that `int` is 32-bit? I think we may want to replace `int` in CSVParser with `int32_t`. So for instance, the lines
```cpp
typedef ParserFactoryReg<uint32_t, int> Reg32int;
typedef ParserFactoryReg<uint64_t, int> Reg64int;
```
would become
```cpp
typedef ParserFactoryReg<uint32_t, int32_t> Reg32int32;
typedef ParserFactoryReg<uint64_t, int32_t> Reg64int32;
```

haojin2: Good catch, I'll make the change and update my PR.

haojin2:
@hcho3 should be good for merge

PR #418, opened 2018-06-27
Fix two bugs in unittest's CMakeLists.txt

JiadingGai:
1. FindGTest.cmake defines a variable called GTEST_LIBRARIES, *not*
GTEST_LIBRARY.

2. Add an include_directories statement to search for include files from
an existing GTest installation.


hcho3:
Thanks!

PR #420, opened 2018-07-02
Modernize lookup logic for OpenSSL SSL and Crypto libraries

hcho3:
Now we require CMake 3.2 instead of 2.8.7.


hcho3:
Before this patch, I had trouble compiling `dmlc-core` with S3 support on my Macbook.

PR #425, opened 2018-07-16
Travis CI build is broken

cclauss:
Even a PR does not change any code will fail the Travis CI build.  See #429 for an attempt to fix.


cclauss:
Fixed in #429

PR #432, opened 2018-07-17
Normalized build for MinGW-w64

KOLANICH:


tqchen:
Unfortunately, this will depend on the availability of cmake. For now, we do want dmlc-core to be header only and not depending on cmake when possible

hcho3:
@tqchen Do we have commitment to support systems without CMake?

tqchen:
there are always cases where we want things like single file packing(use a single source file to compile a module) even if we only use cmake. It is still a good idea to have explicit definitions on the headers(while allowing to resort to non-defaults if they are defined by flag) to support such cases

hcho3:
I still like the idea of detecting the existence of fopen64. Would it be sufficient to include a detection logic in Makefile?

tqchen:
It is possible to do so in CMake, and we just need to add additional -DDMLC_USE_FOPEN64 accordingly

hcho3:
So do we ask MinGW/Cygwin users to use CMake?

KOLANICH:
>It is possible to do so in CMake, and we just need to add additional -DDMLC_USE_FOPEN64 accordingly


This would require detecting and setting this macrodef in all the dependencies. Creating a config is precerred because we don't need to modify build code when we wanna include these headers.

tqchen:
Unfortunately, do we have a requirement of amalgamation and being able to use header-only libraries, detecting things using CMake and set it on xgboost will likely resolve the XGBoost's problem, so i would recommend we go with that path

tqchen:
There are very useful changes make this in this PR can be used to update CMake to run the detection in xgboost's cmake

```
check_symbol_exists(fopen64 stdio.h FOPEN_64_PRESENT)
if(FOPEN_64_PRESENT)
   add_definitions(-DDMLC_USE_FOPEN64=0)
endif
```

KOLANICH:
@tqchen this breaks incapsulation :( I guess even keeping this lib header-only doesn't mean these headers shouldn't be machine-generated.

hcho3:
@KOLANICH I'm not personally familiar with the config files in CMake. How does machine generating `build_config.h` provide better encapsulation? All applications that include `dmlc-core` will now have to include `dmlc/build_config.h`, which to me doesn't look better than including `dmlc/base.h` (which happens indirectly via other dmlc headers).

KOLANICH:
>How does machine generating build_config.h provide better encapsulation?

The build system of the code dependent on dmlc-core doesn't have to set the macrodefs set in the `build_config.h`

tqchen:
I think if we are to rethink things from scratch, maybe we will like to support generated header when there is an install command in cmake, but when things are built in the source, we should not make the assumption that header is being generated.

The main reason is again the simple amalgamation usecase, sometimes we want to allow the user to compile a single file, which includes all the dependencies on the embedded platform and in that case we cannot make the assumption that user call cmake first

tqchen:
So the amalgamation use cases are pretty simple, see https://github.com/dmlc/xgboost/blob/master/amalgamation/xgboost-all0.cc

This is used to build xgboost R package, a user just has to build it with a single command line.

What I am proposing, is always default to ship the files with the default macro values(when value is not defined), as we did before. So users don't have to do anything to do direct amalgamation compile(in certain cases add a predefined macro).

When we do want to ship things in command like make install, we use bash script to amend the content of the default value

hcho3:
@KOLANICH @tqchen I think generated header has merits. In particular, it should help fix build issues with MinGW/Cygwin. For amalgamation use case, we can ship with a pre-built `build_config.h`  with sane default values of the macros.

tqchen:
I am fine with it, as long as build_config.h is shipped in the source, and we allow cmake to override it

KOLANICH:
So, are you merging it?

src/io/single_file_split.h
```diff
@@ -14,9 +14,7 @@
 #include <string>
 #include <algorithm>
 
-#if defined(__FreeBSD__) && DMLC_USE_FOPEN64
-#define fopen64 std::fopen
-#endif
+#include <dmlc/build_config.h>
```
hcho3: Can you move this line to line 10 (before all other headers)?

hcho3:
@KOLANICH Can you add a default `build_config.h` to the source control? Then we'll merge it.

tqchen:
Please add a build_config.h that have the default options to the source. So that amalgamation will work insource without invoking cmake. This is unfortunately a  case we have to support

KOLANICH:
Done it, but CI build for OSX fails, because cmake is not used. Here is the code https://github.com/KOLANICH/hashcat/blame/cmake_build/.travis.yml#L16 to install CMake into travis (needs some fixes), but keep in mind that it takes several minutes.

include/dmlc/build_config.h
```diff
@@ -0,0 +1,6 @@
+#define FOPEN_64_PRESENT
```
hcho3: You should remove this line, since by default FOPEN_64_PRESENT would be 0.

hcho3:
@KOLANICH No need to worry about that. See my comment in `build_config.h`.

KOLANICH:
Done.

hcho3:
Let's exclude `build_config.h` from lint by modifying

https://github.com/dmlc/dmlc-core/blob/14409c983feaa075829f291d5234ba28ff1ef72d/Makefile#L12

to

```
NOLINT_FILES = --exclude_path include/dmlc/concurrentqueue.h include/dmlc/blockingconcurrentqueue.h include/dmlc/build_config.h
``

**Comment by hcho3** [comment]
@tqchen @KOLANICH It looks like we need to decide what the default should be. I see two alternatives:

1. Set `FOPEN_64_PRESENT=0` by default. With this option, we'd need to install CMake in Linux test, since the GNU stdio.h already contains fopen64.
2. Set `FOPEN_64_PRESENT=1` by default. With this option, we'd need to install CMake in Mac test, since the stdio.h in Mac does not contain fopen64.

**Comment by tqchen** [comment]
We can still keep the old detection macros and set the default accordingly while allowing cmake to override it.

**Comment by KOLANICH** [comment]
>have apt quickly install CMake.

Cmake in trusty repos is very old. Not sure if it is useful for us.

>We can still keep the old detection macros and set the default accordingly while allowing cmake to override it.

We can.

**Comment by hcho3** [comment]
@KOLANICH I'd like to do something about MinGW support (and `fopen64` detection). Can I take over this pull request?

**Comment by KOLANICH** [comment]
@hcho3 done

**Comment by hcho3** [comment]
@KOLANICH @tqchen I think we arrived at a happy medium: default `build_config.h` has some logic for people not running CMake, and you can run CMake to overwrite `build_config.h`.

File: include/dmlc/build_config.h [hunk 3]
```diff
@@ -0,0 +1,8 @@
+/* default detection logic for fopen64; may be overriden by CMake */
```
tqchen: add include guard as in other DMLC header files

tqchen:
I have one final comment and after changes are being made this can be merged.

Makefile
```diff
@@ -9,7 +9,7 @@ endif
 include $(config)
 include make/dmlc.mk
 
-NOLINT_FILES = --exclude_path include/dmlc/concurrentqueue.h include/dmlc/blockingconcurrentqueue.h
+NOLINT_FILES = --exclude_path include/dmlc/concurrentqueue.h include/dmlc/blockingconcurrentqueue.h include/dmlc/build_config.h
```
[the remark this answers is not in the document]
hcho3: Done

include/dmlc/build_config.h
hcho3: Done

CMakeLists.txt
```diff
@@ -57,6 +61,10 @@ endif()
 # Older stdc++ enable c++11 items
 add_definitions(-D__USE_XOPEN2K8)
 
+check_symbol_exists(fopen64 stdio.h FOPEN_64_PRESENT)
+message(STATUS "${CMAKE_LOCAL}/build_config.h.in -> ${INCLUDE_DMLC_DIR}/build_config.h")
```
larroy: Why?

hcho3: @larroy Some systems like MinGW have fopen() and fopen64(), and we want to choose fopen64() to use 64-bit offsets

larroy: Thanks for your response, I didn't think you would read it. The problem with this change is that build_config.h is recreated on build, which is marking the 3rdparty submodule dirty in mxnet everytime we build. Is this build_config.h changed by CMake neccessary? I would like to avoid making the repository dirty on build.

Thanks.

hcho3: @larroy Originally, I had planned to have CMake create build_config.h on the fly. The header file contains special macros that are specific to platforms, e.g. existence of fopen64(). The issue was that not everyone uses CMake; many projects using dmlc-core uses GNU Make instead. So I created build_config.h containing the "default logic". Users of GNU Make will use the header as it is. On the other hand, CMake users will overwrite the header, to reflect the platform-specific details that CMake detects.

hcho3: @larroy One solution would be to add build_config.h to .gitignore.

larroy: @hcho3 Thanks for the explanation. Another option is to generate the file with both systems as a configure step. That last one seems more clean to me, so do I understand correctly that the header is there under version control because of Make builds?

larroy: Isn't it already in ignore?  https://github.com/dmlc/dmlc-core/blob/master/.gitignore#L49

hcho3: @larroy Yes, that's correct. Keep in mind that not every user of GNU Make uses Autotools ("configure")

hcho3: @larroy That's weird, maybe path is off?

larroy: I think it won't work if the file is in the repo:

I tried locally.

https://stackoverflow.com/questions/1818895/keep-ignored-files-out-of-git-status

hcho3: Got it. I'm open to suggestion as to how to better accommodate both CMake and GNU Make.

hcho3: For now, you should run
```
git update-index --assume-unchanged include/dmlc/build_config.h
```
from [this page](https://wildlyinaccurate.com/git-ignore-changes-in-already-tracked-files/)

PR #434, opened 2018-07-20
Fix three problems

TccccD:
Fixed three issues:
1. Add "&& s3_region_ == "us-east-1"" to the judgment， s3_region Should be defined in the environment variable ;
2. Add SHA256 when reading data, because the previous contributor modified the write data to add SHA256, but the signed headers bug occurred when reading the data;
3. Add the timeout wait and the header error wait. When reading the data, each epoch will be init request again. But if the server breaks off, the whole training will break off. 
Add the timeout connection wait and the header error wait to avoid this. problem.


TccccD:
@tqchen

PR #435, opened 2018-07-24
Re-add s3_region_ == us-east-1

TccccD:
The previous contributor deleted this option, resulting in the inability to use the s3_region_ set in the environment variable.

### DIFF
--- a/src/io/s3_filesys.cc
+++ b/src/io/s3_filesys.cc
@@ -706,7 +706,7 @@ void ReadStream::InitRequest(size_t begin_bytes,
   std::string canonical_uri;
   CHECK_EQ(path_.name.front(), '/');
   CHECK_NE(path_.host.front(), '/');
-  if (path_.host.find('.', 0) == std::string::npos) {
+  if (path_.host.find('.', 0) == std::string::npos && s3_region_ == "us-east-1") {
     // use virtual host style if no period in host
     canonical_uri = URIEncode(path_.name, false);
     canonical_headers["host"] = path_.host + ".s3.amazonaws.com";
@@ -883,7 +883,7 @@ void WriteStream::Run(const std::string &method,
   std::string canonical_uri;
   std::ostringstream sauth, sdate, stoken, surl, scontent;
   std::ostringstream rheader, rdata;
-  if (path_.host.find('.', 0) == std::string::npos) {
+  if (path_.host.find('.', 0) == std::string::npos && s3_region_ == "us-east-1") {
     canonical_uri = URIEncode(path_.name, false);
     canonical_headers["host"] = path_.host + ".s3.amazonaws.com";
     surl << "https://" << path_.host << ".s3.amazonaws.com"
@@ -1037,7 +1037,7 @@ void S3FileSystem::ListObjects(const URI &path, std::vector<FileInfo> *out_list)
             "&prefix=" + URIEncode(std::string{RemoveBeginSlash(path.name)});
     }
 
-    if (path.host.find('.', 0) == std::string::npos) {
+    if (path.host.find('.', 0) == std::string::npos && s3_region_ == "us-east-1") {
       // use virtual host style if no period in host
       canonical_uri = "/";
       canonical_headers["host"] = path.host + ".s3.amazonaws.com";


TccccD:
@tqchen

TccccD:
Please check this pr, or give me some suggestions. @tqchen ，thanks！

tqchen:
Checking s3_region specific to us-east-1 seems to be pretty restrictive, can you elaborate why this check breaks things and if we need another environment variable to enable the non-AWS related behavior?

tqchen:
cc @rahul003 can you also take a look

TccccD:
In S3FileSystem::S3FileSystem(), :

> const char *region = getenv("S3_REGION");
  if (region == NULL) {
    LOG(WARNING) << "No AWS Region set, using default region us-east-1";
    s3_region_ = "us-east-1";
  } else if (strcmp(region, "") == 0) {
    LOG(WARNING) << "AWS Region was set to empty string, using default region us-east-1";
    s3_region_ = "us-east-1";
  } else {
    s3_region_ = region;
  }

So, s3_region_  is defined  "us-east-1" or others, 
in surl : 

>  if (path.host.find('.', 0) == std::string::npos && s3_region_ == "us-east-1") {
      // use virtual host style if no period in host
      canonical_uri = "/";
      canonical_headers["host"] = path.host + ".s3.amazonaws.com";
      surl << "https://" << path.host << ".s3.amazonaws.com"
           << "/?delimiter=/&prefix=" << RemoveBeginSlash(path.name);
    } else {
      canonical_uri = URIEncode("/" + path.host + "/", false);
      canonical_headers["host"] = s3_endpoint_;
      surl << "https://" << s3_endpoint_ << "/" << path.host << "/?delimiter=/&prefix="
           << RemoveBeginSlash(path.name);
    }

we define s3_region_  and s3_endpoint_ in the environment variable, But if there is no such judgment statement， s3_endpoint_ will not work， and surl will be deferent too。
@tqchen

src/io/s3_filesys.cc
```diff
@@ -1028,7 +1028,7 @@ void S3FileSystem::ListObjects(const URI &path, std::vector<FileInfo> *out_list)
     std::string canonical_uri;
     std::string canonical_querystring;
 
-    AddDefaultCanonicalHeaders(&canonical_headers, curr_time, s3_session_token_, payload, false);
+    AddDefaultCanonicalHeaders(&canonical_headers, curr_time, s3_session_token_, payload, true);
```
rahul003: Isn't payload empty when reading? Is this required? What scenario did you run into which failed?

rahul003:
Hi @TccccD ,
Not sure I understand the issue here. I remember testing the different cases. 

Do you want to use the endpoint style URL in a region which is not us-east-1? Could you give me an indication of how your URL looks, and why the current code doesn't work for you?

TccccD:
> static void AddDefaultCanonicalHeaders(std::map<std::string, std::string> *canonical_headers,
                                       const time_t &curr_time,
                                       const std::string &s3_session_token,
                                       const std::string &data,
                                       bool addDataHash = false) {
  (*canonical_headers)["x-amz-date"] = GetDateISO8601(curr_time);
  if (s3_session_token != "") {
    (*canonical_headers)["x-amz-security-token"] = s3_session_token;
  }
  if (addDataHash) {
    (*canonical_headers)["x-amz-content-sha256"] = SHA256Hex(data);
  }
}

if this set false, canonical_headers will not have x-amz-content-sha256 option,  but in BuildRequestHeaders:

> static void BuildRequestHeaders(std::ostringstream& sauth,
                                std::ostringstream& sdate,
                                std::ostringstream& stoken,
                                std::ostringstream& scontent,
                                const time_t& curr_time,
                                const std::string& s3_access_id,
                                const std::string& s3_region,
                                const std::string& s3_session_token,
                                const std::map<std::string, std::string>& canonical_headers,
                                const std::string& signature,
                                const std::string& payload) {
  sauth << "Authorization: AWS4-HMAC-SHA256 ";
  sauth << "Credential=" << s3_access_id << "/" << GetCredentialScope(curr_time, s3_region) << ",";
  sauth << "SignedHeaders=" << GetSignedHeaders(canonical_headers) << ",";
  sauth << "Signature=" << signature;
  sdate << "x-amz-date: " << GetDateISO8601(curr_time);
  stoken << "x-amz-security-token: " << s3_session_token;
  scontent << "x-amz-content-sha256: " << SHA256Hex(payload);
}

it assigned value to x-amz-content-sha256， It will throw Signed headers error.

TccccD:
i set environment variable：
export S3_REGION=southchina
export S3_ENDPOINT=9.171.32.130:5443

I expect the url to be like https://9.171.32.130:5443/obs-lpf/?delimiter=/&prefix=data/flower

but i get https://obs-lpf.s3.amazonaws.com/?delimiter=/&prefix=data/flower

Due to the lack of “s3_region_ == us-east-1”

@rahul003

rahul003:
@TccccD Could you try specifying the region and endpoint as per 
https://docs.aws.amazon.com/general/latest/gr/rande.html
```

China (Beijing) | cn-north-1 | Valid endpoint name for this region:s3.cn-north-1.amazonaws.com.cn | cn-north-1 | HTTP and HTTPS | Version 4 only
China (Ningxia) | cn-northwest-1 | Valid endpoint name for this region:s3.cn-northwest-1.amazonaws.com.cn | cn-northwest-1 | HTTP and HTTPS | Version 4 only
```

TccccD:
@rahul003 ,But in the block that defines surl，

>   if (path_.host.find('.', 0) == std::string::npos ) {
    // use virtual host style if no period in host
    canonical_uri = URIEncode(path_.name, false);
    canonical_headers["host"] = path_.host + ".s3.amazonaws.com";
    surl << "https://" << path_.host << ".s3.amazonaws.com" << '/' << RemoveBeginSlash(path_.name);
  } else {
    canonical_uri = URIEncode("/" + path_.host + path_.name, false);
    canonical_headers["host"] = s3_endpoint_;
    surl << "https://" << s3_endpoint_ << '/' << path_.host << '/'
         << RemoveBeginSlash(path_.name);
  }

it is “.s3.amazonaws.com”，which is very restrictive，so this should re-add “s3_region_ == "us-east-1".

src/io/s3_filesys.cc
```diff
@@ -883,7 +883,7 @@ void WriteStream::Run(const std::string &method,
   std::string canonical_uri;
   std::ostringstream sauth, sdate, stoken, surl, scontent;
   std::ostringstream rheader, rdata;
-  if (path_.host.find('.', 0) == std::string::npos) {
+  if (path_.host.find('.', 0) == std::string::npos && s3_region_ == "us-east-1") {
     canonical_uri = URIEncode(path_.name, false);
     canonical_headers["host"] = path_.host + ".s3.amazonaws.com";
```
rahul003: I got your point @TccccD . But I think the solution here is to change ".s3.amazonaws.com" to ".s3."+s3_region_+".amazonaws.com" . That would work for any region instead of adding this check to make it only for us-east-1.

rahul003: Please be sure to specify s3_region as per the AWS region names like cn-north-1

rahul003: Ah okay. Missed that. 

I had already added a function in the code which determines the end point for a s3 region. Could you please use that to set the host.

`path_.host + s3::getEndpoint(s3_region_)`

TccccD: I don't understand what you mean, where can I use this function?
@rahul003

rahul003:
```
  if (path_.host.find('.', 0) == std::string::npos) {
    canonical_uri = URIEncode(path_.name, false);
    canonical_headers["host"] = path_.host + "." + s3::getEndpoint(s3_region_);
    surl << "https://" << canonical_headers["host"] 
          << path_.name << "?" << GetQueryMultipart(params, false);
   } else {
     canonical_uri = URIEncode("/" + path_.host + path_.name, false);
     canonical_headers["host"] = s3_endpoint_;
     surl << "https://" << s3_endpoint_ << "/" << path_.host
          << path_.name << "?" << GetQueryMultipart(params, false);
   }
```

PR #436, opened 2018-07-24
fix bug when use path_imgidx in ImageRecordIter

solin319:
When use ImageRecordIter read data set on S3, use std::ifstream to read idx file on S3 will cause segment fault.
Can we change it to dmlc::Stream created by fs->Open.
@tqchen


tqchen:
We can, but we will need to close it or wrap the dmlc::Stream in std::unique_ptr

solin319:
We have wrapped the dmlc::Stream in std::unique_ptr.

solin319:
@tqchen

tqchen:
Thanks for the contribution, this is merged

PR #438, opened 2018-07-27
Use separated hostfile for workers and servers

xcgoner:
For ssh tracker, we add the support of using different hostfiles for the worker group and server group. Thus, the servers and workers are not necessarily co-located, which can avoid exhausting cpu resources.


xcgoner:
Oh, sorry, I didn't noticed that. Actually I cannot find the document as you mentioned. Could you please indicate where to find such document?

rahul003:
https://mxnet.incubator.apache.org/faq/distributed_training.html Under using launch.py options section.

xcgoner:
Got it. Thanks.

PR #444, opened 2018-08-02
Add S3_IS_AWS env and fixed non-AWS behavior

TccccD:
1. If do not add new environment variables, it is difficult to determine whether it is a non-AWS user, so an S3_IS_AWS is added here;
2. Fixed an issue with s3.amazonaws.com and "s3."+ region_name + ".amazonaws.com.cn"
3. "<?xml version="1.0" encoding="UTF-8" standalone="yes"?><Error><Code>InvalidArgument</Code><Message>Request has x-amz-content-sha256 header but signed headers not include it.</Message><RequestId>" error occurred while reading data.

### DIFF
--- a/src/io/s3_filesys.cc
+++ b/src/io/s3_filesys.cc
@@ -1180,7 +1180,6 @@ S3FileSystem::S3FileSystem() {
   } else {
     s3_isAWS_ = false;
   }
-
   if (region == NULL) {
     LOG(WARNING) << "No AWS Region set, using default region us-east-1";
     s3_region_ = "us-east-1";
--- a/src/io/s3_filesys.cc
+++ b/src/io/s3_filesys.cc
@@ -671,9 +671,11 @@ class ReadStream : public CURLReadStreamBase {
              const std::string &s3_region,
              const std::string &s3_endpoint,
              const bool s3_verify_ssl,
+             const bool s3_isAWS,
              size_t file_size)
       : path_(path), s3_id_(s3_id), s3_key_(s3_key), s3_session_token_(s3_session_token),
-         s3_region_(s3_region), s3_endpoint_(s3_endpoint), s3_verify_ssl_(s3_verify_ssl) {
+         s3_region_(s3_region), s3_endpoint_(s3_endpoint), s3_verify_ssl_(s3_verify_ssl),
+         s3_isAWS_(s3_isAWS){
     this->expect_file_size_ = file_size;
   }
   virtual ~ReadStream(void) {}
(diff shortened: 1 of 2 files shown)


rahul003:
I didn't realize there can be non-AWS S3!

tqchen:
apparently S3 is an open standard, @rahul003 can you check and explicitly approve?

src/io/s3_filesys.cc
```diff
@@ -699,18 +701,19 @@ void ReadStream::InitRequest(size_t begin_bytes,
...
-  AddDefaultCanonicalHeaders(&canonical_headers, curr_time, s3_session_token_, payload, false);
+  AddDefaultCanonicalHeaders(&canonical_headers, curr_time, s3_session_token_, payload, true);
   std::ostringstream sauth, sdate, stoken, surl, scontent, srange;
   std::ostringstream result;
   std::string canonical_querystring;
   std::string canonical_uri;
   CHECK_EQ(path_.name.front(), '/');
   CHECK_NE(path_.host.front(), '/');
-  if (path_.host.find('.', 0) == std::string::npos) {
+  if (s3_isAWS_) {
```
rahul003: Please make this `s3_is_AWS_ && path_.host.find('.', 0) == std::string::npos)` because there are cases when you would use second case even with AWS.

src/io/s3_filesys.cc
```diff
@@ -883,10 +887,10 @@ void WriteStream::Run(const std::string &method,
   std::string canonical_uri;
   std::ostringstream sauth, sdate, stoken, surl, scontent;
   std::ostringstream rheader, rdata;
-  if (path_.host.find('.', 0) == std::string::npos) {
+  if (s3_isAWS_) {
```
rahul003: Same here

src/io/s3_filesys.cc
```diff
@@ -1162,13 +1167,19 @@ S3FileSystem::S3FileSystem() {
     region = getenv("AWS_REGION");
   }
 
+
```
rahul003: nit: remove this line

src/io/s3_filesys.h
```diff
@@ -81,6 +81,7 @@ class S3FileSystem : public FileSystem {
   std::string s3_endpoint_;
   /*! \brief S3 verify ssl*/
   bool s3_verify_ssl_;
+  bool s3_isAWS_;
```
rahul003: could you rename this to s3_is_aws_

TccccD:
ok， please check it， thanks！

PR #445, opened 2018-08-06
Add OMPException class and use it for Text Parser

anirudh2290:
Adding a class for OMPException since this needs to be handled in TextParser in dmlc-core and also the ImageDetRecordIOParser, ImageRecordIOParser and other parsers that maybe added in future in MXNet.

### DIFF
--- a/src/data/text_parser.h
+++ b/src/data/text_parser.h
@@ -9,6 +9,7 @@
 
 #include <dmlc/data.h>
 #include <dmlc/omp.h>
+#include <dmlc/common.h>
 #include <thread>
 #include <mutex>
 #include <vector>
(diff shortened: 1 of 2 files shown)


hcho3:
@anirudh2290 Can you use this for `#pragma omp for` loops as well?

anirudh2290:
@hcho3 yes you should be able to use it and i am using it in the related PR for mxnet https://github.com/apache/incubator-mxnet/pull/12051/files

anirudh2290:
@hcho3 does that answer your question ? is this PR good to go ?

hcho3:
@anirudh2290 Let me review it today.

src/data/text_parser.h
```diff
@@ -120,34 +119,24 @@ inline bool TextParserBase<IndexType, DType>::FillData(
...
-      }
-      ParseBlock(pbegin, pend, &(*data)[tid]);
-    } catch (dmlc::Error& ex) {
-      {
-        std::lock_guard<std::mutex> lock(mutex_exception_);
-        if (!parser_exception_) {
-          parser_exception_ = std::current_exception();
-        }
-      }
+  omp_exc_.Run([=] {
```
hcho3: I'm a bit wary about the default value capture here. For example, is it cheap to copy `chunk`?

I think we should stick to capture by reference `[&]`, since there is no danger of the lambda outliving the captured objects.

anirudh2290: good point. you are right it should be by reference.  i fixed it at other places in mxnet code but missed it here.

PR #452, opened 2018-08-31
fix bug in reading overlapped data in the distributed mode

htgeis:
I encounter an error when using the distributed XGBoost(see here for detail https://discuss.xgboost.ai/t/errors-in-distributed-training/186). 
The reason is caused by the reading and parsing data logic here instead of the data itself. 
It happened when reading the last chunk in a worker. The detail is that we reserve extra newlines between remaining files but these spaces may be used to read overlapped data which should be handle in next worker. In this case, offset_curr_ will be greater than offset_end_ and it will bypass the FindLastRecordBegin which could correct the boundary. 
One of the debug log I have printed for it is shown that the end lines of data in ParseBlock(libsvm_parser.cc) is incomplete.
e.g. "1.0 1:0 2.0 .....  1113:0.0 1114:0.0 1115:0.0 1116:0.0 \n 1.0 1"
And it would check failed when the last feature contains key only.

Besides, the previous code also has risk of ArrayIndexOutOfBounds Exception since we add num_extra_eol to the  size(https://github.com/dmlc/dmlc-core/blob/master/src/io/input_split_base.cc#L185) and it may exceed the length of Chunk if we have too many files left.

### DIFF
--- a/src/io/input_split_base.cc
+++ b/src/io/input_split_base.cc
@@ -226,7 +226,7 @@ bool InputSplitBase::ReadChunk(void *buf, size_t *size) {
                             max_size - olen);
   nread += olen;
   if (nread == 0) return false;
-  if (nread == olen) { // add extra newline to handle files with NOEOL
+  if (nread == olen) {  // add extra newline to handle files with NOEOL
     char *bufptr = reinterpret_cast<char*>(buf);
     bufptr[nread] = '\n';
     nread++;
--- a/src/io/input_split_base.cc
+++ b/src/io/input_split_base.cc
@@ -180,9 +180,7 @@ size_t InputSplitBase::Read(void *ptr, size_t size) {
   }
   if (offset_begin_ >= offset_end_) return 0;
   if (offset_curr_ +  size > offset_end_) {
-    // allow for extra newlines between files
-    const size_t num_extra_eol = files_.size() - file_ptr_ - 1;
-    size = offset_end_ - offset_curr_ + num_extra_eol;
+    size = offset_end_ - offset_curr_;
   }
   if (size == 0) return 0;
   size_t nleft = size;
@@ -228,20 +226,16 @@ bool InputSplitBase::ReadChunk(void *buf, size_t *size) {
                             max_size - olen);
   nread += olen;
   if (nread == 0) return false;
-  if (nread != max_size) {
-    *size = nread;
-    return true;
-  } else {
-    const char *bptr = reinterpret_cast<const char*>(buf);
-    // return the last position where a record starts
-    const char *bend = this->FindLastRecordBegin(bptr, bptr + max_size);
-    *size = bend - bptr;
-    overflow_.resize(max_size - *size);
-    if (overflow_.length() != 0) {
-      std::memcpy(BeginPtr(overflow_), bend, overflow_.length());
-    }
-    return true;
+
+  const char *bptr = reinterpret_cast<const char*>(buf);
+  // return the last position where a record starts
+  const char *bend = this->FindLastRecordBegin(bptr, bptr + nread);
+  *size = bend - bptr;
+  overflow_.resize(nread - *size);
+  if (overflow_.length() != 0) {
+    std::memcpy(BeginPtr(overflow_), bend, overflow_.length());
   }
+  return true;
 }
 
 bool InputSplitBase::Chunk::Load(InputSplitBase *split, size_t buffer_size) {
--- a/src/io/input_split_base.cc
+++ b/src/io/input_split_base.cc
@@ -226,6 +226,11 @@ bool InputSplitBase::ReadChunk(void *buf, size_t *size) {
                             max_size - olen);
   nread += olen;
   if (nread == 0) return false;
+  if (nread == olen) { // add extra newline to handle files with NOEOL
+    char *bufptr = reinterpret_cast<char*>(buf);
+    bufptr[nread] = '\n';
+    nread++;
+  }
 
   const char *bptr = reinterpret_cast<const char*>(buf);
   // return the last position where a record starts
--- a/test/unittest/unittest_inputsplit.cc
+++ b/test/unittest/unittest_inputsplit.cc
@@ -184,3 +184,45 @@ TEST(InputSplit, test_split_libsvm) {
   ASSERT_EQ(num_row, 5U);
   ASSERT_EQ(num_col, 125U);
 }
+
+TEST(InputSplit, test_split_libsvm_distributed) {
+  {
+    /* Create a test case for partitioned libsvm */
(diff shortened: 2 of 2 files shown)


hcho3:
I introduced extra newlines between files to handle partitioned LIBSVM files with NOEOL. See #389.
Let us check if your fix properly handles partitioned LIBSVM with NOEOL.

htgeis:
I test the case in #389 and it works well. I know the reason about adding extra newlines, which make it possible  to read the whole remaining content in one iteration. But it could caused the issue that read more data which should belong to the next worker and it may exceeds array bounds also. Thus, I don't add extra newlines but I do some record boundary check to fix it.

hcho3:
Did you find a way to handle files that doesn't end with \n (NOEOL)?

hcho3:
Also, would it be possible to come up with a small case to reveal the bug? We should add a regression test so that the same bug doesn't occur again.

htgeis:
Thanks for your reminder. I add the check for NOEOL. It can works well now. Besides, I have prepared the data to reveal the bug.
Minimal reproduction:
Create five text files named test/test_0.libsvm, test/test_1.libsvm, test/test_2.libsvm, test/test_3.libsvm, test/test_4.libsvm.
test/test_0.libsvm has the content as follows

> 1 3:1 10:1 11:1 21:1 30:1 34:1 36:1 40:1 41:1 53:1 58:1 65:1 69:1 77:1 86:1 88:1 92:1 95:1 102:1 105:1 117:1 124:1
1 3:1 10:1 11:1 21:1 30:1 34:1 36:1 40:1 41:1 53:1 58:1 65:1 69:1 77:1 86:1 88:1 92:1 95:1 102:1 105:1 117:1 124:1
1 3:1 10:1 11:1 21:1 30:1 34:1 36:1 40:1 41:1 53:1 58:1 65:1 69:1 77:1 86:1 88:1 92:1 95:1 102:1 105:1 117:1 124:1
1 3:1 10:1 11:1 21:1 30:1 34:1 36:1 40:1 41:1 53:1 58:1 65:1 69:1 77:1 86:1 88:1 92:1 95:1 102:1 105:1 117:1 124:1
1 3:1 10:1 11:1 21:1 30:1 34:1 36:1 40:1 41:1 53:1 58:1 65:1 69:1 77:1 86:1 88:1 92:1 95:1 102:1 105:1 117:1 124:1
1 3:1 10:1 11:1 21:1 30:1 34:1 36:1 40:1 41:1 53:1 58:1 65:1 69:1 77:1 86:1 88:1 92:1 95:1 102:1 105:1 117:1 124:1

and other file has the same content as follows

> 1 3:1 10:1 11:1 21:1 30:1 34:1 36:1 40:1 41:1 53:1 58:1 65:1 69:1 77:1 86:1 88:1 92:1 95:1 102:1 105:1 117:1 124:1

image

Then run distributed XGBoost with two worker and It will fail.

image

hcho3:
Thanks a lot for the example. This should help us add a regression test. Let me look over your pull request this weekend and get back to you.

src/io/input_split_base.cc
```diff
@@ -228,20 +226,21 @@ bool InputSplitBase::ReadChunk(void *buf, size_t *size) {
...
-    const char *bend = this->FindLastRecordBegin(bptr, bptr + max_size);
-    *size = bend - bptr;
-    overflow_.resize(max_size - *size);
-    if (overflow_.length() != 0) {
-      std::memcpy(BeginPtr(overflow_), bend, overflow_.length());
-    }
-    return true;
+  if (nread == olen) {  // add extra newline to handle files with NOEOL
+    char *bufptr = reinterpret_cast<char*>(buf);
+    bufptr[nread] = '\n';
```
hcho3: I don't think we'd want to add `\n` here, since line 194 is already adding one

hcho3: I removed lines 229-233 and ran two unit tests in `test/unittest/unittest_inputsplit.cc`. Both passed.

htgeis: I don't think so. It would get failed when the last the file is NOEOL(that's why it passed your test case) if we remove lines 229-233.
It can be fixed in two ways.
(1) add '\n' just like lines 229-233 shows.
(2) or modify the function FindLastRecordBegin to handle this case.(Maybe it should not return the beginPtr when the data contains one character "\n" only). But I found that it needs to change several files so I choose the Solution One.

hcho3: You should try it on your end too: remove lines 229-233 and run InputSplit unit tests.

htgeis: @hcho3  You could test it by creating the test case with a single file with NOEOL and it will drop into an infinite loop if we remove lines 229-233 here.

hcho3: > creating the test case with a single file with NOEOL and it will drop into an infinite loop if we remove lines 229-233 here.

Is this with two machines?

hcho3: @htgeis Ah got it. I do get infinite loop with a single NOEOL file. I will add another unit test to account for this.

hcho3: @htgeis I added a unit test to test cases where the last file has NOEOL. Let me know what you think.

htgeis:
thanks for your kind review and test cases.

Vikas89:
This change is causing segfault.
Refer here: https://github.com/apache/incubator-mxnet/issues/12800

If I revert the changes in input_split_base.cc, it fixes the problem. I think adding "\n" is not allowed for recordio_split
@hcho3 @htgeis

Vikas89:
My changes in dmlc-core/src/io/input_split_base.cc which fixes the problem

```
+  if (nread != max_size) {
+    *size = nread;
+    return true;
+  } else {
+    const char *bptr = reinterpret_cast<const char*>(buf);
+    // return the last position where a record starts
+    const char *bend = this->FindLastRecordBegin(bptr, bptr + max_size);
+    *size = bend - bptr;
+    overflow_.resize(max_size - *size);
+    if (overflow_.length() != 0) {
+      std::memcpy(BeginPtr(overflow_), bend, overflow_.length());
+    }
+  
+    return true;
+  }
+    /* COMMENTOUT
   if (nread == olen) {  // add extra newline to handle files with NOEOL
     char *bufptr = reinterpret_cast<char*>(buf);
     bufptr[nread] = '\n';
     nread++;
+    LOG(INFO) << " Added extra line to the file";
   }
 
   const char *bptr = reinterpret_cast<const char*>(buf);
@@ -241,6 +258,7 @@ bool InputSplitBase::ReadChunk(void *buf, size_t *size) {
     std::memcpy(BeginPtr(overflow_), bend, overflow_.length());
   }
   return true;
+  */

```

PR #459, opened 2018-10-02
Fix gtest can not find source build issue

wweic:
When I run `make -f scripts/packages.mk gtest`, it failed with error:
````
mv googletest-release-1.7.0 gtest
cd gtest; g++ -Iinclude -pthread -c src/gtest-all.cc -o gtest-all.o; cd ..
src/gtest-all.cc:42:10: fatal error: 'src/gtest.cc' file not found
#include "src/gtest.cc"
         ^~~~~~~~~~~~~~         
````

Reproduced on Mac OS and Ubuntu. It's also an issue for tvm cpp unittest build since it uses this script to install `gtest`. The fix is simple, just add the needed path.


wweic:
@tqchen please help review, should be a quick one.

PR #462, opened 2018-10-07
Fix #456: Enable usleep for non-MSVC on Windows

### LINKED ISSUE #456: usleep is not available on Windows
https://github.com/dmlc/dmlc-core/blob/649be18a8c55c48517861d67158a45dec54992ee/test/unittest/unittest_thread_group.cc#L32

`usleep` is used. It is not available (in fact, it is, see https://github.com/KOLANICH/nixShims4Windows/blob/master/src/time.c#L2 , it uses the undocumented function) on Windows.

You have an own shim, but it is enabled only for MSVC
https://github.com/dmlc/dmlc-core/blob/649be18a8c55c48517861d67158a45dec54992ee/test/unittest/unittest_thread_group.cc#L10

But it is not needed, 1e4 microseconds is 10 milliseconds, whic is in range of [`Sleep`](https://docs.microsoft.com/en-us/windows/desktop/api/synchapi/nf-synchapi-sleep)

hcho3:
Fixes #456 by defining `usleep()` for all Windows targets, not just when Visual Studio is used.


hcho3:
TODO: set up CI for Windows using AppVeyor

PR #469, opened 2018-10-16
Redesign dmlc::TemporaryDirectory to allow recursive deletion

hcho3:
Use `dmlc::io::LocalFileSystem::ListDirectory()` to allow recursive deletion of temporary directories. Now users don't have to call `AddFile()`; it suffices to create files inside the directory.

Limitation: no symbolic links are allowed inside the temporary directory.

### DIFF
--- a/include/dmlc/filesystem.h
+++ b/include/dmlc/filesystem.h
@@ -32,7 +32,8 @@ namespace dmlc {
  * \brief Manager class for temporary directories. Whenever a new
  *        TemporaryDirectory object is constructed, a temporary directory is
  *        created. The directory is deleted when the object is deleted or goes
- *        out of scope.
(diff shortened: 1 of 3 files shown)


hcho3:
@trivialfis Can you take a look if you get a chance?

trivialfis:
@hcho3 LGTM. Thank you.  I added a very tiny test to verify, which is attached in this comment. But I don't think it's important anyway, your change looks good by itself. :)

I don't know anything about Windows,  so I skipped anything related to Windows.  

[0001-Add-small-tests-for-TemporaryDirectory.txt](https://github.com/dmlc/dmlc-core/files/2486298/0001-Add-small-tests-for-TemporaryDirectory.txt)

hcho3:
@trivialfis Thanks. I added a unit test for TemporaryDirectory.

Issue #460, opened 2018-10-06
There should be exportation of include directories in CMake file.


navidR:
After [this line ](https://github.com/dmlc/dmlc-core/blob/master/CMakeLists.txt#L153) there should be a line along this line : 

```
target_include_directories(dmlc PUBLIC include)
```

FYI this is a temporary fix. A better solution would be only using target_include_directories instead of include_directories. I am willing to send a small patch to fix CMake files if you are willing to accept patches.

hcho3:
@navidR 
> After this line there should be a line along this line

Can you describe what kind of problems arise because we don't have `target_include_directories` currently?

navidR:
You are using include_directories. There are a bunch of problems with this. 

First of all, in my case, I am using dmlc-core as a subdirectory in my cmake project. So when you don't use `target_include_directories`,  when I do `target_link_libraries(${my_project} PRIVATE dmlc)` then include include directories of dlmc-core does not get included automatically. Which makes a pain in the back.

Second, this is not the correct way in Modern CMake. As I told you, I am willing to send a patch to modernize your cmake implementation since I am going to use dmlc-core and halideir in one of my projects.

For more information you can look at this links: 
https://stackoverflow.com/questions/31969547/what-is-the-difference-between-include-directories-and-target-include-directorie
https://cliutils.gitlab.io/modern-cmake/

hcho3:
Sure, ability to automatically locate dmlc-core headers sounds nice. You should submit a pull request so that me and other maintainers can review it.

Some considerations:
* There are several projects that uses dmlc-core, such as [apache/incubator-mxnet](https://github.com/apache/incubator-mxnet) and [dmlc/xgboost](https://github.com/dmlc/xgboost). I don't think they follow what you call "Modern CMake" practices, your pull request should not break existing CMake bulids.
* To make this a learning opportunity for me and others: is there significant consensus in C++ community about what constitutes Modern CMake or its merits? Is Modern CMake akin to Modern C++ in terms of adoption and support? I know a bit about what Modern C++ looks like (e.g. don't manage raw pointers; use smart pointers / resource-managing classes), but to me the term Modern CMake is quite new.

navidR:
About the first point, sure. I understand how painful it is to rewrite portions of your build tools. So I will keep it 100% compatible.

[...]

PR #470, opened 2018-10-16
 Enable specifying stacktrace size at runtime.

trivialfis:
@hcho3 
This PR also groups GTest CMake configuration together. Seems to have some conflicts with previous commits.  Let me see if it breaks the tests.  
The change is mostly for https://github.com/dmlc/xgboost/pull/3795 .  With this I don't need the `PROFILING` option.


trivialfis:
Still trying to figure out the best way to profile the memory usage. :)  Sorry for these early PRs.

hcho3:
@trivialfis Don't be sorry. Your contribution to dmlc-core is quite valuable.

hcho3:
@trivialfis Is this PR ready for review?

trivialfis:
@hcho3 Yes. Along with https://github.com/dmlc/xgboost/pull/3795 .  The only problem is we are hesitating to merge it due to:

- We don't know whether nvprof will support this in near future.
- I'm kind of the only one who's running it.

Merging https://github.com/dmlc/xgboost/pull/3795 will be another large chunk of code, we are not sure it's worthy. You can take a look into it if you have the time.
If we decide not to merge profiling, I will extract other changes in this PR and create a new one.

trivialfis:
@hcho3 Thanks.  It's ready for review.

include/dmlc/logging.h
```diff
@@ -319,14 +319,14 @@ inline std::string Demangle(char const *msg_str) {
   return string(msg_str);
 }
 
-inline std::string StackTrace() {
+inline std::string StackTrace(
+    const int stack_size = DMLC_LOG_STACK_TRACE_SIZE) {
```
hcho3: Any reason to make the parameter const? I haven't seen any examples where non-pointer parameters are made const.

hcho3: Also let's make the parameter unsigned, so that negative inputs won't be allowed.

test/unittest/CMakeLists.txt
```diff
@@ -1,22 +1,60 @@
...
 add_executable(${PROJECT_NAME}_unit_tests ${UNIT_TEST_SOURCE})
 set_property(TARGET ${PROJECT_NAME}_unit_tests
-        PROPERTY RUNTIME_OUTPUT_DIRECTORY ${PRIVATE_RUNTIME_DIR})
+  PROPERTY RUNTIME_OUTPUT_DIRECTORY ${PRIVATE_RUNTIME_DIR})
 
-message(STATUS "${CMAKE_CURRENT_SOURCE_DIR}/build_config.h.in -> ${CMAKE_CURRENT_SOURCE_DIR}/build_config.h")
-configure_file("${CMAKE_CURRENT_SOURCE_DIR}/build_config.h.in" "${CMAKE_CURRENT_SOURCE_DIR}/build_config.h")
+message(STATUS "${CMAKE_CURRENT_SOURCE_DIR}/build_config.h.in -> ${CMAKE_CURRENT_BINARY_DIR}/build_config.h")
+configure_file("${CMAKE_CURRENT_SOURCE_DIR}/build_config.h.in" "${CMAKE_CURRENT_BINARY_DIR}/build_config.h")
+target_include_directories(${PROJECT_NAME}_unit_tests PRIVATE ${CMAKE_CURRENT_BINARY_DIR})
```
hcho3: Isn't this duplicate of line 25?

trivialfis: Ops, something went wrong during rebase.  Thanks.

include/dmlc/logging.h
trivialfis: Here `const` is just an habit.

test/unittest/unittest_inputsplit.cc
```diff
@@ -147,7 +147,7 @@ TEST(InputSplit, test_split_libsvm_distributed) {
 #ifdef DMLC_UNIT_TESTS_USE_CMAKE
 /* Don't run the following when CMake is not used */
 
-#include <build_config.h>
+#include "build_config.h"
```
hcho3: Why this change?

trivialfis: There are two "build_config.h".
Just want to make it clearer that the one in the same directory is used.

hcho3: That makes sense. Can you change it to `"./build_config.h"` to make it even clearer?

trivialfis: Done. :)

include/dmlc/logging.h
```diff
@@ -344,7 +344,7 @@ inline std::string demangle(char const* msg_str) {
   return std::string();
 }
 
-inline std::string StackTrace() {
+inline std::string StackTrace(const int stack_size = 0) {
```
hcho3: This parameter should be also size_t

hcho3: Once this is fixed, I'll merge it

trivialfis:
@hcho3 I looked into the source code of backward-cpp.  It supports a wide ranges of "backend" libraries.  It's already very messy to achieve what it does,  if we were to add this into dmlc-core, I think we better limit the support to `libdw` only.

For build option, we can utilize CMake's `CMAKE_CXX_FLAGS_DEBUG` variable.

But let me focus on https://github.com/dmlc/xgboost/pull/3825 first, I'm a little bit tied up to that currently.  I will come back to this.

hcho3:
@trivialfis I don't think we should add backward-cpp to dmlc-core, since it is quite messy.

hcho3:
@trivialfis Actually, I take back what I said, since all errors get directed to `LOG(FATAL)`, which is part of dmlc-core.

Anyway, let's hold it for now, since we have other things to worry about. Also, I think we want to "modernize" the CMake build before we attempt to do something fancy like backward-cpp. See discussion at #460 about Modern CMake practices.

PR #471, opened 2018-10-18
Do not add newline for RecordIO InputSplit

hcho3:
Adding newlines caused apache/incubator-mxnet#12800.

TODO: Add a test case


hcho3:
Adding (optional) cURL to download a file from the Internet during a C++ test.

hcho3:
Turns out cURL was an overfill. CMake has facility to download files.

test/unittest/CMakeLists.txt
```diff
@@ -8,9 +8,24 @@ file(GLOB_RECURSE UNIT_TEST_SOURCE "*.cc")
 add_executable(${PROJECT_NAME}_unit_tests ${UNIT_TEST_SOURCE})
 set_property(TARGET ${PROJECT_NAME}_unit_tests
         PROPERTY RUNTIME_OUTPUT_DIRECTORY ${PRIVATE_RUNTIME_DIR})
+
+# File downloads
+message(STATUS "Downloading http://data.mxnet.io/data/cifar10/cifar10_val.rec...") 
+file(DOWNLOAD "http://data.mxnet.io/data/cifar10/cifar10_val.rec"
```
Vikas89: Are we downloading entire cifar10_val.rec. Since this is unit test, I think this will add unnecessary delay to test. Can we have a minimal test file of few bytes?

src/io/input_split_base.cc
```diff
@@ -191,7 +193,9 @@ size_t InputSplitBase::Read(void *ptr, size_t size) {
     offset_curr_ += n;
     if (nleft == 0) break;
     if (n == 0) {
-      buf[0] = '\n'; ++buf; --nleft;  // insert a newline between files
+      if (is_text_parser) {
+        buf[0] = '\n'; ++buf; --nleft;  // insert a newline between files
```
Vikas89: can you write a comment, why are we inserting \n for text parser

src/io/input_split_base.cc
```diff
@@ -226,12 +230,18 @@ bool InputSplitBase::ReadChunk(void *buf, size_t *size) {
...
   nread += olen;
   if (nread == 0) return false;
-  if (nread == olen) {  // add extra newline to handle files with NOEOL
-    char *bufptr = reinterpret_cast<char*>(buf);
-    bufptr[nread] = '\n';
-    nread++;
+  if (this->IsTextParser()) {
+    if (nread == olen) {  // add extra newline to handle files with NOEOL
+      char *bufptr = reinterpret_cast<char*>(buf);
+      bufptr[nread] = '\n';
```
Vikas89: comment explaining why \b here again. In what case this \n will be helpful compared to previous \n

test/unittest/.gitignore
```diff
@@ -1 +1,3 @@
 dmlc_unittest
+build_config.h
+cifar10_val.rec
```
Vikas89: is this required, now ?

src/io/input_split_base.cc
hcho3: I'll add a link to the issue that gave rise to it.

test/unittest/.gitignore
hcho3: I'll remove `cifar10_val.rec`.

src/io/input_split_base.cc
hcho3: I'll add a link to the issue that gave rise to it.

PR #265, opened 2017-06-06
Add build support for cygwin

sunnijie:
1. use `std::fopen` instead of `fopen64` under cygwin
2. change `--c++0x` to `--gnu++11` to fix error: "fileno was not declared in this scope"
3. turn off `DMLC_LOG_STACK_TRACE` under cygwin to fix error: "execinfo.h: No such file or directory"

PR #474, opened 2018-10-26
Add Windows tests for VS 2017 + MinGW-w64 + Cygwin

hcho3:
Run the unit tests for more compilers on Windows so as to catch compilation problems.

Related: #265


hcho3:
@trivialfis FYI, with this change, backtrace is now available for Mac OSX if CMake is used. (It turns out `execinfo.h` is present on recent versions of Mac OSX.)

trivialfis:
Cool.  But that's a lot of meta code in there.  I don't know what's happening in between these platform dependent macros. :(

hcho3:
@trivialfis The idea is that you can explicitly check whether a function / header exists on the user's system. For example, `DMLC_EXECINFO_H` macro  tells us if `execinfo.h` is available or not.

Yeah, unfortunately, things can get messy with platform dependent code. Such is life :(

PR #475, opened 2018-10-29
[Experiment] add user to config breath of tree based all reduce sync

chenqin:
default is binary tree, user could fine tune their settings if needed.


chenqin:
screen shot 2018-10-29 at 11 08 36

CodingCat:
do we have test covering this?

PR #476, opened 2018-10-30
FreeBSD support in endian.h

mallniya:


include/dmlc/endian.h
```diff
@@ -8,7 +8,7 @@
 
 #include "./base.h"
 
-#if defined(__APPLE__) || defined(_WIN32)
+#if defined(__APPLE__) || defined(__FreeBSD__) || defined(_WIN32)
```
hcho3: FreeBSD is not necessarily little endian. See https://lists.freebsd.org/pipermail/freebsd-questions/2008-February/167918.html.
To determine endianness of FreeBSD, we can include `<sys/endian.h>`

hcho3: FYI: `<sys/endian.h>` is available on all BSDs, according to this doc: http://man7.org/linux/man-pages/man3/endian.3.html

hcho3: @mallniya In the future, I will add FreeBSD to CI tests to ensure that dmlc-core compiles on FreeBSD. For now, let me update this PR to include better endian detection logic, as found in https://www.boost.org/doc/libs/1_54_0/boost/detail/endian.hpp

hcho3: @mallniya Can you try compiling with my changes?

mallniya: @hcho3 yes, compilation succeed.

hcho3: Cool!

tqchen:
I would say let us keep endian simple, and ask users to explicit declare it

hcho3:
@tqchen Any concrete suggestion about how to move forward with FreeBSD? Should we ask FreeBSD users to use CMake?

tqchen:
Yes, i think use cmake is a good suggestion

tqchen:
If we can add a quick macro support BSD, I am fine with that

mallniya:
That's really good news, thank you!

PR #480, opened 2018-11-08
Using /bin/bash in tracker

mirocody:
This change is to support remote ssh when doing distributed training with mxnet and DLAMI https://github.com/awslabs/deeplearning-cfn. 

We need to use /bin/bash to remotely run''source activate <conda_env>'' for distributed training.

### DIFF
--- a/tracker/dmlc_tracker/tracker.py
+++ b/tracker/dmlc_tracker/tracker.py
@@ -363,7 +363,7 @@ class PSTracker(object):
         for k, v in envs.items():
             env[k] = str(v)
         self.thread = Thread(
-            target=(lambda: subprocess.check_call(self.cmd, env=env, shell=True)), args=())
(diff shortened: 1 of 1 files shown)


Vikas89:
==> We need to use /bin/bash to remotely run''source activate <conda_env>'' for distributed training.

Can you explain why /bin/bash is needed and other shell doesn't work?

mirocody:
See comments in 
https://github.com/bokeh/bokeh/pull/2490
https://github.com/conda/conda/issues/1512
Conda might not support dash and other shell.
I also tried to run '. activate <conda_env>', but it cannot activate other conda environment successfully, which cause 
``import mxnet as mx`` failed with ``no module found`` error(which indicates it does not activate mxnet conda environment).

gautamkmr:
travis-ci has failed to download few files. Can you make sure PR is rebased and resubmit here?

PR #490, opened 2018-12-07
Disable header_guard in lint.py

grwlf:
For error discussion, see
https://discuss.tvm.ai/t/problem-running-task-lint-sh-locally/1285
https://github.com/nodejs/node/issues/2693


tqchen:
I don't think this is the right approach. We do usually need to have a good convention of the header guard. I will follow up in the discuss thread

PR #494, opened 2018-12-19
Destructor undefined

lihaozhehw:
The purpose of the destructor is to do the corresponding cleanup when the class is released, so that the coder does not forget the cleanup of the resource, so if there is a constructor, there must be a destructor.


tqchen:
Default destructor is defined when there is no decalration, so it is fine to not define explicit destructors

PR #495, opened 2019-01-03
Add install target.

trivialfis:
Export `dmlc` target in CMake


trivialfis:
@hcho3

trivialfis:
Just found out this has some duplication to #257 .  But this one adds export target.

trivialfis:
@hcho3 I removed `filesys.h` and moved everything in it into `io.h` for the following reasons:
- Originally `src/io/filesys.h` is included by `include/dmlc/filesystem.h`, which effectively making it part of interface.  But we can't install such a file in `CMakeLists.txt` since it's located in `src/io` directory.
- Simply moving this file into `include/dmlc` is confusing, what's the reason for having both `filesystem.h` and `filesys.h`?
- Moving it into `io.h` instead of `filesystem.h` is because originally `filesys.h` is located in the sub-directory of `src` named `io`.

I added a simple documentation of using dmlc-core with CMake, could you help reviewing this?

PR #497, opened 2019-01-31
Avoid double-free in LogCheckError

alexeyr:
The default copy constructor of `LogCheckError` could lead to double-free errors. An alternative to this fix would be to define the constructor and assigment operator (and move constructor?).

### DIFF
--- a/include/dmlc/logging.h
+++ b/include/dmlc/logging.h
@@ -68,8 +68,14 @@ class LogCheckError {
  public:
   LogCheckError() : str(nullptr) {}
   explicit LogCheckError(const std::string& str_) : str(new std::string(str_)) {}
+  LogCheckError(const LogCheckError& other) = delete;
+  LogCheckError(LogCheckError&& other) : str(other.str) {
+    other.str = nullptr;
+  }
   ~LogCheckError() { if (str != nullptr) delete str; }
-  operator bool() {return str != nullptr; }
+  operator bool() const { return str != nullptr; }
+  LogCheckError& operator=(const LogCheckError& other) = delete;
+  LogCheckError& operator=(LogCheckError&& other) = delete;
   std::string* str;
 };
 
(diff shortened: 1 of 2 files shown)


include/dmlc/parameter.h
```diff
@@ -310,7 +310,7 @@ namespace parameter {
 class FieldAccessEntry {
  public:
   FieldAccessEntry()
-      : has_default_(false) {}
+      : has_default_(false), index_(0) {}
```
hcho3: Why are we adding a new field `index_` here?

tqchen:
I believe the reason that a ptr is used is to minimize the cost and allow constant folding. so we'd better keep it as it is

include/dmlc/logging.h
```diff
@@ -66,11 +66,10 @@ inline void InitLogging(const char*) {
...
 class LogCheckError {
  public:
-  LogCheckError() : str(nullptr) {}
-  explicit LogCheckError(const std::string& str_) : str(new std::string(str_)) {}
-  ~LogCheckError() { if (str != nullptr) delete str; }
-  operator bool() {return str != nullptr; }
-  std::string* str;
+  LogCheckError() : str() {}
+  explicit LogCheckError(const std::string& str_) : str(str_) {}
+  operator bool() const { return !str.empty(); }
```
hcho3: I think an empty value is different from a missing value. Maybe we should use `std::unique_ptr<std::string>` to store the reference to the temporary string?

alexeyr: In general yes, but here it's used for an error message, which should never be empty if present. In particular, it's only called as `LogCheckError("Error.")` or `LogCheckError(os.str())` with non-empty `os`.

include/dmlc/parameter.h
alexeyr: We aren't, just initializing it.

alexeyr:
@tqchen The new version keeps the member as pointer.

include/dmlc/logging.h
alexeyr: Anyway, this is now changed according to @tqchen's comment, probably `unique_ptr` would be as undesirable as my initial solution.

PR #501, opened 2019-02-05
Follow-up for #495: fix config path

hcho3:
When dmlc-core is used as a git submodule of another project, we should not use `CMAKE_SOURCE_DIR`.


hcho3:
@trivialfis

PR #506, opened 2019-02-28
Add noexcept in DataIter, InputSplit

arcadiaphy:
This PR is created in addition to [https://github.com/apache/incubator-mxnet/pull/14223](https://github.com/apache/incubator-mxnet/pull/14223) and [https://github.com/dmlc/mshadow/pull/369](https://github.com/dmlc/mshadow/pull/369) to address memory issues in mxnet exit.

When TensorContainer's destructor becomes noexcept(false), classes that reference it also become noexcept(false) implicitly in destruction.

If inherited class's destructor is noexcept(false) and its base class has a virtual destructor, then base class's destructor should also be added noexcept(false) to successfully compile.

### DIFF
--- a/include/dmlc/common.h
+++ b/include/dmlc/common.h
@@ -10,6 +10,7 @@
 #include <string>
 #include <sstream>
 #include <mutex>
+#include <utility>
 #include "./logging.h"
 
 namespace dmlc {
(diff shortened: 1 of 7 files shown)


szha:
@arcadiaphy thanks for the fix. would you mind also look into the lint errors?

arcadiaphy:
@szha lint is ok

PR #512, opened 2019-03-08
[rabit harden] remove rabit_num_trail from dmlc-core

chenqin:
Follow up refactor based on comments after https://github.com/dmlc/dmlc-core/pull/510. 

- Introduce **--local-num-attempt** to config total number of times local tracker could restart workers.(for obvious reasons, this config serves little meaning in distributed environment)
- Pass updated **DMLC_NUM_ATTEMPT** value to mocked rabit worker as number of failed tails local tracker executed in current job.

This pr served a prerequisite of reenable all rabit C++ tests (following pr here https://github.com/dmlc/rabit/pull/81)

### DIFF
--- a/tracker/dmlc_tracker/local.py
+++ b/tracker/dmlc_tracker/local.py
@@ -21,6 +21,7 @@ def exec_cmd(cmd, role, taskid, pass_env):
     env['DMLC_TASK_ID'] = str(taskid)
     env['DMLC_ROLE'] = role
     env['DMLC_JOB_CLUSTER'] = 'local'
+
     # keep use DMLC_NUM_ATTEMPT for backward compatible reason
     num_retry = env.get('DMLC_NUM_ATTEMPT', 0)
 
--- a/tracker/dmlc_tracker/local.py
+++ b/tracker/dmlc_tracker/local.py
@@ -8,12 +8,13 @@ import subprocess
 import logging
 from threading import Thread
 from . import tracker
+import pdb
 
 def exec_cmd(cmd, role, taskid, pass_env):
     """Execute the command line command."""
     if cmd[0].find('/') == -1 and os.path.exists(cmd[0]) and os.name != 'nt':
         cmd[0] = './' + cmd[0]
-    cmd = ' '.join(cmd)
+    cmdline = ' '.join(cmd)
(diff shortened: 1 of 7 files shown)


tracker/dmlc_tracker/local.py
```diff
@@ -43,7 +46,7 @@ def exec_cmd(cmd, role, taskid, pass_env):
             if num_retry >= 0:
                 # failure trail increase by 1 and restart failed worker
                 for arg in cmd:
-                    if arg.startswith('rabit_num_trial'):
+                    if arg.startswith('DMLC_NUM_ATTEMPT'):
```
CodingCat: it's very weird to have this in local.py, this is essentially a param only for test, and particularly only for any test requiring to indicate how many times the process has been started

CodingCat: the expected solution is to make dmlc-submit as spark-submit, it takes several parameters for its own, like --num-executors, and anything following the path to a jar should be taken as the application-specific params, and they are delegate to the application itself to process

chenqin:
@CodingCat @hcho3

chenqin:
passed xgb integration tests
https://travis-ci.org/chenqin/xgboost/builds/505366624

tracker/dmlc_tracker/local.py
```diff
@@ -21,12 +21,11 @@ def exec_cmd(cmd, role, taskid, pass_env):
...
     env['DMLC_JOB_CLUSTER'] = 'local'
-    num_retry = env.get('DMLC_NUM_ATTEMPT', 0)
 
-    #overwrite default num of retry with commandline value
-    for param in cmd:
-        if param.startswith('DMLC_NUM_ATTEMPT'):
-            num_retry = int(param.split('=')[1])
+    # backward compatibility
+    num_retry = env.get('DMLC_NUM_ATTEMPT', num_attempt)
+    num_trail = 0
```
szha: trial?

chenqin: fixed.

chenqin:
^ @szha can we merge this to master https://github.com/dmlc/rabit/issues/82 is point to this fix.

chenqin:
Thanks!

PR #516, opened 2019-03-12
add android to endian detection

eqy:


include/dmlc/endian.h
```diff
@@ -12,7 +12,7 @@
   // If compiled with CMake, use CMake's endian detection logic
   #define DMLC_LITTLE_ENDIAN DMLC_CMAKE_LITTLE_ENDIAN
 #else
-  #if defined(__APPLE__) || defined(_WIN32)
+  #if defined(__APPLE__) || defined(_WIN32) || defined(__ANDROID__)
```
tqchen: actually ```__ANDROID__``` do have endian.h , so you can use below endian detection which is more robust

PR #519, opened 2019-03-18
Fix misprint

ruslo:


szha:
@ruslo thanks for the patch!

PR #520, opened 2019-03-23
make REGISTER thread safe and idempotent

CodingCat:
it will facilitate dynamic registering new function entries with multi threads

for example the under-developing feature in https://github.com/dmlc/xgboost/pull/4280


CodingCat:
@hcho3

PR #521, opened 2019-03-23
Add noexcept in DataIter, InputSplit

hcho3:
This PR applies #506 to the master branch. After this PR is merged, we should be able to address #507, i.e. MXNet will be able to point to the latest dmlc-core again.


hcho3:
@szha

PR #523, opened 2019-04-10
Support RISC-V cross compilation

nhynes:
This allows TVM to build using a RISC-V GCC toolchain (https://github.com/rv8-io/musl-riscv-toolchain)


hcho3:
Is using CMake an option for you?

nhynes:
I am using CMake?

hcho3:
@nhynes If you are using CMake all the way, CMake is able to detect endianness as well as the presence of backtrace libs:
https://github.com/dmlc/dmlc-core/blob/3ffea8694adf9c0363f9abbf162dc0e4a45b22c5/CMakeLists.txt#L83-L101

The point is moot, however, since you said you have to use plain Make. I will approve this pull request.

PR #524, opened 2019-04-16
[rabit harden] address subprocess exit deadlock

chenqin:
https://bugs.python.org/issue20318 Caused rabit test hang and timeout from time to time. As python community has no plan to fix in 2.7. We should use process to replace thread to run each rabit worker.

### DIFF
--- a/tracker/dmlc_tracker/local.py
+++ b/tracker/dmlc_tracker/local.py
@@ -68,7 +68,8 @@ def submit(args):
                 role = 'worker'
             else:
                 role = 'server'
-            # the problem of
+
+            # the problem of https://bugs.python.org/issue20318
             # using thread means branch from same process shared by all threads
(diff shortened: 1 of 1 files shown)


hcho3:
@chenqin Given that Python 2.x would be EOL end of this year, wouldn't using Python 3 solve the issue?

CodingCat:
yes, and it would be nice if you can help to test xgboost's tracker in python3

chenqin:
python3 seems work on my ubuntu and osx, just confirm, so are we agree to move foward and end support of python2 in next release?

hcho3:
@chenqin @CodingCat I opened a new thread to get votes from the community: #4379. Related: dmlc/tvm#2994

CodingCat:
too many PRs opened and it becomes hard to track which one we should focus on 

close this one as we are moving to py3

PR #525, opened 2019-04-17
[python3] use python3 dmlc tracker in rabit tests

chenqin:
upgrade to dmlc-submit to python3 as alternative to https://github.com/dmlc/dmlc-core/pull/524

### DIFF
--- a/scripts/lint.py
+++ b/scripts/lint.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 # pylint: disable=protected-access, unused-variable, locally-disabled, len-as-condition
 """Lint helper to generate lint summary of source.
 
--- a/tracker/dmlc-submit
+++ b/tracker/dmlc-submit
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 import sys
 import os
 curr_path = os.path.dirname(os.path.abspath(os.path.expanduser(__file__)))
--- a/tracker/dmlc_tracker/kubernetes.py
+++ b/tracker/dmlc_tracker/kubernetes.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 """
 DMLC submission script by kubernetes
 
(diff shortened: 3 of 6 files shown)


chenqin:
https://github.com/dmlc/xgboost/issues/4379

CodingCat:
how we verify python3 can make rabit test pass?

chenqin:
I tested in my local OS X and Linux, to have Travis chi working we need to get either this or deadlock or in and git update sub module

CodingCat:
as the common practice like before, we should point rabit to the private branch having python 3 to verify  

and 

there are still bunch of places in the code base using python instead of python3, we'd fix in one shot instead of leaving code base in an inconsistent status using py2 and py3 simultaneously

chenqin:
This is getting convoluted, if we need to fix rabit problem do we need to wait all xgboost upgrade to python3? Otherwise, we should merge https://github.com/dmlc/dmlc-core/pull/524 and keep upgrade python3 in parallel

CodingCat:
>  if we need to fix rabit problem do we need to wait all xgboost upgrade to python3?

make a simple search, https://github.com/dmlc/dmlc-core/search?l=Python&q=python  

> Otherwise

otherwise against what? 

> we should merge #524 

alternatives as highlighted

image

chenqin:
524 assume we support python2 and python3 
525 meant to drop support of python2 and assume deadlock issue go away with python2

unless we can assume https://github.com/dmlc/xgboost/tree/master/python-package/xgboost works with python3, python3 upgrade might be more than we expected.  Ultimately, my impression is we should get away all python2 packages from pip install and move to pip3.

CodingCat:
> unless we can assume https://github.com/dmlc/xgboost/tree/master/python-package/xgboost works with python3, python3 upgrade might be more than we expected. 

it is not related to what we are talking about here, and if you have any concern or you think you find some cases XGBoost cannot move forward to py3, please stated in https://github.com/dmlc/xgboost/issues/4379

> Ultimately, my impression is we should get away all python2 packages from pip install and move to pip3.

yes, that's why I said, if you want to upgrade to python 3 in this code base, you'd upgrade everything appeared in https://github.com/dmlc/dmlc-core/search?l=Python&q=python

hcho3:
@chenqin FYI, XGBoost has been tested with Python 3 for a good while.

PR #526, opened 2019-04-25
Allow selection of build tools for gtest-all

anshu-dasgupta:
This is needed to build successfully on some systems.

In general, this will allow more flexibility in building gtest. Specifically, for our use case, it allows gtest to be compiled with clang. This avoids conflicts between libc++ and libstdc++ because of the way our build systems are configured.


anshu-dasgupta:
Note that the check on the original commit showed one test failing (see https://travis-ci.org/dmlc/dmlc-core/builds/524598201) I suspect that this was a test/timing problem. I don't see how this commit can cause that test failure. The retriggered Travis CI build passed.

PR #529, opened 2019-05-01
stringstream default behaviour will lose precision

FrozenGene:
Currently, when we accept  float / double values, we will use stringstream. However, the default behaviour of stringstream will lose precision of float / double.

For example:

```cpp
int main() {
    double d = 0.1234567890987654;
    std::stringstream ss;
    ss << d;
    std::cout << ss.str() << "\n";
    ss.str("");
    ss << std::setprecision(std::numeric_limits<double>::max_digits10) << d;
    std::cout << ss.str() << "\n";
}
```

output:
0.123457
0.1234567890987654

We could see we will lose precision if we don't set precision. This PR fix this issue. 

Background: It is found in TVM when we pass Python's double value to TVM's NNVM, then get back to Python, finally we get incorrect result.

### DIFF
--- a/include/dmlc/parameter.h
+++ b/include/dmlc/parameter.h
@@ -1014,7 +1014,7 @@ class FieldEntry<float> : public FieldEntryNumeric<FieldEntry<float>, float> {
     }
   }
 
-  protected:
+ protected:
   // print the value
   virtual void PrintValue(std::ostream &os, float value) const {  // NOLINT(*)
     os << std::setprecision(std::numeric_limits<float>::max_digits10) << value;
@@ -1053,7 +1053,7 @@ class FieldEntry<double>
     }
   }
 
-  protected:
+ protected:
   // print the value
   virtual void PrintValue(std::ostream &os, double value) const {  // NOLINT(*)
     os << std::setprecision(std::numeric_limits<double>::max_digits10) << value;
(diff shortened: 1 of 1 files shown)


FrozenGene:
@tqchen, I am not sure who I should ask for reviewer. Could you help to @ and assign? Thanks.

include/dmlc/parameter.h
```diff
@@ -1012,6 +1013,12 @@ class FieldEntry<float> : public FieldEntryNumeric<FieldEntry<float>, float> {
       throw dmlc::ParamError(os.str());
     }
   }
+
+ protected:
+  // print the value
+  virtual void PrintValue(std::ostream &os, float value) const {  // NOLINT(*)
+    os << std::setprecision(std::numeric_limits<float>::max_digits10) << value;
```
hcho3: This will change the internal state of `os`. Is `os` used somewhere else?

hcho3: Side-effect free alternative:
```suggestion
    char buf[512];
    sprintf(buf, "%.*g", std::numeric_limits<float>::max_digits10, value);
    os << buf;
```

FrozenGene: `os` is passed by `GetStringValue`, which is a local parameter. It just accept value and return `os.str()`, doesn’t do anything else. So I think `setprecision` is also ok.

hcho3:
Merged. Thanks!

PR #537, opened 2019-06-13
support set/get configuration in tracker

chenqin:
xgb native checkpoint recovery were broken because restart worker run allreduce before loadcheckpoint.
if tracker support set/get shared configuration, we can have native rabit recovery running.

https://github.com/dmlc/xgboost/issues/4250#issuecomment-501023606


chenqin:
looks like last pr https://github.com/dmlc/dmlc-core/commit/fbe142b267a8edd1f1188fa2140d88f7ae308661 cause ci failure. 
https://travis-ci.org/chenqin/dmlc-core/jobs/545091181

PR #540, opened 2019-07-11
Declare the type name of `int64_t`

wkcn:
In MXNet, `dim_t` is the type of some parameters, like [`mx.nd.eye`](https://github.com/apache/incubator-mxnet/blob/master/src/operator/tensor/init_op.h#L84). 

However, `dim_t` couldn't be parsed into a string, since `dim_t` is the alias of `int64_t`, which is not registed by dmlc-core type-traits.

In this PR, I declare the type name of `int64_t`.


wkcn:
@hcho3 
Thank you!

I have a question whether it is necessary to declare other types, such as `char`, `short`?

hcho3:
@wkcn I have no idea. I'm assuming this PR is sufficient for your use case though?

wkcn:
@hcho3 Yes, this PR is sufficient for my use case.

PR #541, opened 2019-07-16
Simplify Endian detection logic

hcho3:
It turns out we don't really need endian detection at compile-time. Performing endian detection at run-time is much simpler, and we no longer need neither CMake nor a massive chain of conditional macros.

Update. We need the endian detection function to be a legal `constexpr`, so that it can be used inside a template.


hcho3:
Hmm, apparently, Visual Studio 2013 doesn't have proper support for `constexpr`. Since Windows is exclusively little-endian, I might just add a special macro just for Windows.

Note that the failing test `unittest_gest` in Travis has nothing to do with this pull request; it is also failing for #540 as well.

rongou:
@hcho3 getting compiler warnings:
```
/home/rou/src/xgboost/dmlc-core/include/dmlc/./././endian.h(21): warning: integer conversion resulted in truncation
```

hcho3:
@rongou Which compiler are you using?

hcho3:
I tried running the function `__dmlc_is_system_little_endian` on an emulated SPARC machine (with QEMU), and it returns true when it should actually return false (since SPARC is big endian).

rongou:
@hcho3 I think all the warnings come from the cuda compiler.
```shell
$ nvcc --version
nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2019 NVIDIA Corporation
Built on Wed_Apr_24_19:10:27_PDT_2019
Cuda compilation tools, release 10.1, V10.1.168
$ gcc --version
gcc (Ubuntu 7.4.0-1ubuntu1~18.04.1) 7.4.0
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

PR #544, opened 2019-07-19
Set up CI for big-endian target

hcho3:
#543 raised the need for a big-endian target in the CI tests. For now, use s390x Docker container (with QEMU emulation) as a new big-endian target.

* Add Dockerfile and convenience scripts to run tests inside s390x Docker seamlessly
* Remove RecordIO from unit test for time being, as it's not endian-portable.
* Add a note about RecordIO in README
* Fix build_config.h, so that `DMLC_CMAKE_LITTLE_ENDIAN` is always present when CMake is used.
* Fix Config parser, since the use of `char` to store EOF causes undefined behavior on big-endian machines.
* Add a unit test for `DMLC_LITTLE_ENDIAN` macro

cc @larroy


hcho3:
Build results: https://travis-ci.org/dmlc/dmlc-core/jobs/561192570

nhynes:
cc @trivialfis

PR #546, opened 2019-07-25
fix mpich env detection in mpi tracker

chenqin:
failed to run mpich version of MPI locally, should be misc fix


chenqin:
@hcho3

PR #547, opened 2019-07-26
Enable building with /MD on CMake/Windows

BarclayII:
New Visual Studio projects by default link with dynamic CRT (i.e. with `/MD` option).  However, dmlc-core always build with static CRT (i.e. with `/MT` option), so linking dmlc-core with other projects built with `/MD` throws an error.

This PR adds a CMake option that allows the project to be built with `/MD`.  Similar option is also seen in [Google test](https://github.com/google/googletest/tree/master/googletest#visual-studio-dynamic-vs-static-runtimes) and [TVM](https://github.com/dmlc/tvm/blob/master/CMakeLists.txt#L82).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -108,15 +108,17 @@ configure_file("${CMAKE_LOCAL}/build_config.h.in" "${INCLUDE_DMLC_DIR}/build_con
 # compiler flags
 if(MSVC)
   add_definitions(-DDMLC_USE_CXX11)
-  #if(NOT BUILD_SHARED_LIBS AND NOT DMLC_FORCE_SHARED_CRT)
-  foreach(flag_var
-        CMAKE_CXX_FLAGS CMAKE_CXX_FLAGS_DEBUG CMAKE_CXX_FLAGS_RELEASE
-        CMAKE_CXX_FLAGS_MINSIZEREL CMAKE_CXX_FLAGS_RELWITHDEBINFO)
-    if(${flag_var} MATCHES "/MD")
-      string(REGEX REPLACE "/MD" "/MT" ${flag_var} "${${flag_var}}")
-    endif(${flag_var} MATCHES "/MD")
-  endforeach(flag_var)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -107,13 +108,15 @@ configure_file("${CMAKE_LOCAL}/build_config.h.in" "${INCLUDE_DMLC_DIR}/build_con
...
 if(MSVC)
   add_definitions(-DDMLC_USE_CXX11)
-  foreach(flag_var
-        CMAKE_CXX_FLAGS CMAKE_CXX_FLAGS_DEBUG CMAKE_CXX_FLAGS_RELEASE
-        CMAKE_CXX_FLAGS_MINSIZEREL CMAKE_CXX_FLAGS_RELWITHDEBINFO)
-    if(${flag_var} MATCHES "/MD")
-      string(REGEX REPLACE "/MD" "/MT" ${flag_var} "${${flag_var}}")
-    endif(${flag_var} MATCHES "/MD")
-  endforeach(flag_var)
+  if(NOT BUILD_SHARED_LIBS AND NOT DMLC_FORCE_SHARED_CRT)
```
jermainewang: Where is `BUILD_SHARED_LIBS` defined?

BarclayII: It's a [CMake global option](https://cmake.org/cmake/help/v3.15/variable/BUILD_SHARED_LIBS.html).

jermainewang:
Hi @hcho3 , we found the CI sometimes will fail (non-deterministically) on the `ThreadedIter.dmlc_exception` test case. However, the change here has nothing to do with that part. We suspect there are some concurrency bugs there. Are you aware of this problem?

hcho3:
@jermainewang Hmm we should definitely take a look.

hcho3:
@jermainewang I created #550 to keep track.

VoVAllen:
Can we merge this? @jermainewang

Issue #553, opened 2019-08-01
Perform thorough audit to address concurrency problems


hcho3:
When I compiled dmlc-core with Thread Sanitizer, I got a **massive heap** of concurrency errors: 
[log.txt](https://github.com/dmlc/dmlc-core/files/3458811/log.txt).

This is bad news: multi-threaded code in dmlc-core (threaded iterator, thread groups, concurrent queues etc) have concurrency issues such as data races.

Related: dmlc/dmlc-core#550, dmlc/dmlc-core#505

cc @trivialfis

hcho3:
Excellent resource: [C++ Concurrency in Action](https://www.amazon.com/dp/1617294691/ref=cm_sw_em_r_mt_dp_U_Jv2qDbG0A2SPK)

hcho3:
Potential deadlock in lock-free queue:
```
[----------] 2 tests from Lockfree
[ RUN      ] Lockfree.ConcurrentQueue
==================
WARNING: ThreadSanitizer: lock-order-inversion (potential deadlock) (pid=12822)
  Cycle in lock order graph: M2609 (0x7b2400002f68) => M210 (0x7ffd8acc5a58) => M2609

  Mutex M210 acquired here while holding mutex M2609 in main thread:
    #0 pthread_mutex_lock <null> (libtsan.so.0+0x3fadb)
    #1 __gthread_mutex_lock /usr/include/x86_64-linux-gnu/c++/7/bits/gthr-default.h:748 (dmlc_unit_tests+0x6c2bf)
    #2 __gthread_recursive_mutex_lock /usr/include/x86_64-linux-gnu/c++/7/bits/gthr-default.h:810 (dmlc_unit_tests+0x6c345)
    #3 std::recursive_mutex::lock() <null> (dmlc_unit_tests+0x6fd48)
    #4 std::unique_lock<std::recursive_mutex>::lock() <null> (dmlc_unit_tests+0x75067)
    #5 std::unique_lock<std::recursive_mutex>::unique_lock(std::recursive_mutex&) /usr/include/c++/7/bits/std_mutex.h:197 (dmlc_unit_tests+0x72d04)
    #6 dmlc::ThreadGroup::add_thread(std::shared_ptr<dmlc::ThreadGroup::Thread>) /home/ubuntu/dmlc-core/include/dmlc/thread_group.h:370 (dmlc_unit_tests+0x71732)
[...]

PR #554, opened 2019-08-02
Fix data race and thread leak in ThreadedIter

### LINKED ISSUE #550: Fix memory error in unit test ThreadedIter.dmlc_exception
[ RUN      ] ThreadedIter.dmlc_exception
dmlc_unittest(63607,0x7fffb7c29380) malloc: *** error for object 0x7fff7f4e6770: pointer being freed was not allocated
*** set a breakpoint in malloc_error_break to debug

I suspect this is a race condition because it happens only once in a while.

hcho3:
Fixes #550

* **Prevent "thread leak"**: Wrap `std::thread*` with `std::unique_ptr<>` and `ScopedThread` so that we can automatically `join()` a thread whenever an exception is thrown. Do not use raw pointer to store the thread, as it's not exception-safe.

* **Eliminate data races in producer signals**: Make `producer_sig_processed_` and `producer_sig_` atomic because
  1) Multiple threads access these variables;
  2) The variables are often used without a lock.
 
  Furthermore, we use `.store(*, std::memory_order_release)` and `.load(std::memory_order_acquire)` so that a write from one thread is guaranteed to be visible from another thread. See [Release-Acquire Ordering](https://en.cppreference.com/w/cpp/atomic/memory_order#Release-Acquire_ordering).

* Enable building dmlc-core with Sanitizer. For example, to use Thread Sanitizer, run:
```
cmake .. -DGOOGLE_TEST=ON -DDMLC_USE_SANITIZER=ON \
         -DDMLC_ENABLED_SANITIZERS="thread" -DCMAKE_BUILD_TYPE=Debug ..
```

cc @larroy @jermainewang

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,6 +21,10 @@ dmlccore_option(USE_CXX14_IF_AVAILABLE "Build with C++14 if the compiler support
 dmlccore_option(GOOGLE_TEST "Build google tests" OFF)
 dmlccore_option(INSTALL_DOCUMENTATION "Install documentation" OFF)
 dmlccore_option(DMLC_FORCE_SHARED_CRT "Build with dynamic CRT on Windows (/MD)" OFF)
+dmlccore_option(DMLC_USE_SANITIZER "Use santizer flags; to specify a custom path for sanitizers, set this variable a value that's not ON or OFF" OFF)
+set(DMLC_ENABLED_SANITIZERS "address" "leak" CACHE STRING
+  "Semicolon separated list of sanitizer names. E.g 'address;leak'. Supported sanitizers are
+  address, leak and thread.")
 
 # include path
 set(INCLUDE_ROOT "${CMAKE_CURRENT_SOURCE_DIR}/include")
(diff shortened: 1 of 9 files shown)


jermainewang:
Consider enable sanitizer in the CI after these fixes?

hcho3:
@jermainewang There's still a long way to go until we resolve all errors reported by the Sanitizer: #553

CMakeLists.txt
```diff
@@ -20,13 +20,27 @@ dmlccore_option(USE_OPENMP "Build with OpenMP" ON)
 dmlccore_option(USE_CXX14_IF_AVAILABLE "Build with C++14 if the compiler supports it" OFF)
 dmlccore_option(GOOGLE_TEST "Build google tests" OFF)
 dmlccore_option(INSTALL_DOCUMENTATION "Install documentation" OFF)
+dmlccore_option(USE_SANITIZER "Use santizer flags" OFF)
+dmlccore_option(SANITIZER_PATH "Path to sanitizes." "")
```
junrushao1994: Some future discussion here. TVM has already adopted a style that unifies USE_XXX and XXX_PATH: if USE_XXX is “on”, it searches automatically; otherwise if it is not “off”, it means a manually set path.

Another thing, though I am not confident (because I am not an expert in CMake), I would love to mention is, what if a upstream project includes DMLC core, but they have some options overlap? For example, both of them have USE_LLVM so they conflict. Our practice is to add prefix to USE_XXX, like PROJECT_NAME_USE_XXX to avoid using the same name. Is there any better way to deal with this?

hcho3: I will add a prefix so that `USE_SANITIZER` doesn't have a name clash

trivialfis:
We can fix them one by one?

hcho3:
@trivialfis Yes, I plan to merge this PR pretty soon, and then tackle remaining concurrency issues in subsequent PRs.

hcho3:
@trivialfis Since there are outstanding concurrency issues, I'm afraid we are not yet able to run sanitizer.

hcho3:
@trivialfis For now, I added the sanitizer test and marked it non-blocking.

PR #556, opened 2019-08-21
Allow update of parameters.

trivialfis:


trivialfis:
@hcho3 @RAMitchell Will give a demo in XGBoost after being ready.

trivialfis:
ping @hcho3

test/unittest/unittest_param.cc
```diff
@@ -20,7 +21,8 @@ TEST(Parameter, parsing_float) {
   std::map<std::string, std::string> kwargs;
 
   kwargs["float_param"] = "0";
-  ASSERT_NO_THROW(param.Init(kwargs));
+  param.Init(kwargs);
+  // ASSERT_NO_THROW(param.Init(kwargs));
```
hcho3: Why are we commenting out this line?

trivialfis: Oops.

test/unittest/unittest_param.cc
```diff
@@ -157,3 +159,30 @@ TEST(Parameter, parsing_float) {
...
+  param.float_param = 0.02;
+  param.UpdateAllowUnknown(Args{{"float_param", "0.02"},
+                                {"foo", "bar"}}, &changed);
+  ASSERT_FALSE(changed);
+
+  param.UpdateAllowUnknown(Args{{"foo", "bar"}}, &changed);
+  ASSERT_FALSE(changed);
+
+  param.UpdateAllowUnknown(Args{{"double_param", "0.13"},
+                                {"foo", "bar"}}, &changed);
```
hcho3: What is the expected value of `changed` after this statement?

trivialfis:
Fixed.

include/dmlc/parameter.h
```diff
@@ -575,12 +642,23 @@ class FieldEntryBase : public FieldAccessEntry {
...
-  virtual std::string GetStringValue(void *head) const {
+  bool Same(void* head, std::string const& value) const override {
+    DType old = this->Get(head);
+    DType now;
+    std::istringstream is(value);
+    is >> now;
+    // don't require = operator
+    bool is_same = std::equal(
+        reinterpret_cast<char*>(&now), reinterpret_cast<char*>(&now) + sizeof(now),
+        reinterpret_cast<char*>(&old));
```
hcho3: Are `old` and `new` same only when `old` and `new` have same set of bits? How about `old=-0.0` and `new=0.0`? What is the reason for avoiding the use of `=` operator?

hcho3: According to IEEE 754, `-0.0` and `0.0` are distinct values (with different bits) and at the same time `-0.0 == 0.0` should evaluate to `true`.

trivialfis:
There's a vector in XGB tree param.

We might be able to use type traits to test the concept `EqualityComparable` defined in C++ 20.  Let me try that.

hcho3:
Got it. For now, bit equality should be good.

trivialfis:
@hcho3 Can we merge it?  I will update `dmlc-core` in XGBoost.

PR #558, opened 2019-08-29
Remove throw spec from omp.h for clang

larroy:
This fixes compilation with OMP using the Clang compiler.

### DIFF
--- a/include/dmlc/omp.h
+++ b/include/dmlc/omp.h
@@ -11,7 +11,7 @@
 #include <omp.h>
 #else
 
-#if defined(__ANDROID__)
+#if defined(__ANDROID__) && defined (__clang__)
 #undef __GOMP_NOTHROW
 #define __GOMP_NOTHROW
 #elif defined(__cplusplus)


hcho3:
I suspect the memory error is due to race condition

larroy:
@hcho3 in CI ?

hcho3:
@larroy Yes, the memory error you can see in Travis CI is intermittent

larroy:
https://github.com/dmlc/dmlc-core/pull/560

larroy:
https://github.com/dmlc/dmlc-core/pull/559

larroy:
@hcho3 can we merge this? I need it for an MXNet PR.

hcho3:
Merging this for now. I'll create a new issue about failing tests for dmlc_exception and std_exception.

larroy:
seems it's not failing due to double free anymore.

leezu:
@larroy should this be ` #if defined(__ANDROID__) || defined (__clang__)` instead of ` #if defined(__ANDROID__) && defined (__clang__)`. The second doesn't make much sense to me and in fact doesn't work with clang

larroy:
I think the intention was to fix it for Android only, when using clang, hence &&, don't remember the details. But see that we are compiling with Clang in CI for a while, why did it start failing now?

Issue #538, opened 2019-06-19
yarn-submit.py not runnable using python3


ericangelokim:
Popen().communicate() returns a binary, but the line here tries to split using a str, causing a failure when run with python3.

https://github.com/dmlc/dmlc-core/blob/3943914eed66470bd010df581e29e4dca4f7df6f/tracker/dmlc_tracker/yarn.py#L48

ericaws:
This code works in Python3 as long as the encoding is set to ```'utf-8'```:

```out = out.decode().split('\n')[0].split()```

hcho3:
@ericaws Can you submit a pull request?

ericaws:
I need to add some more code to make this both py2 and py3 compatible

thvasilo:
Running into this issue too when trying to submit XGBoost jobs, especially since the scripts were changed to use python3 by default in https://github.com/dmlc/dmlc-core/pull/525 (ping @chenqin). 

I'm wondering if this would make all yarn submissions to break atm, that's definitely my experience. I'll also note that AWS EMR clusters use an AMI that defaults to python2.7 further complicating things.

chenqin:
upgrade to python3 as alternative to thread bug in python2, if we can have quick fix, we can fix forward .
otherwise we could revert change to python2 and pick https://github.com/dmlc/dmlc-core/pull/524/ instead?

@hcho3 @thvasilo

thvasilo:
I'm under the impression that this is a Python 3 bug, the code works in 2.7
IIRC.
I'm all for phasing out any Python 2.7 code ASAP, but I assume that would
require an RFC.

Regards,
Theodore

--

hcho3:
Is it possible to use Python 3 in AWS EMR?

thvasilo:
Yup, that's what I'm doing (I think, it's kinda hard to tell what gets
called where and how) . If I'm not mistaken however this bug is a Python 3
bug.

--

hcho3:
I think we should revise the YARN tracker to be compatible with Python 3, now that `dmlc-submit` is using Python 3.

ericangelokim:
This can (hackily) be done by checking python version at the beginning of file, and using different API calls throughout the module.

hcho3:
@ericangelokim Let me submit a pull request to revise `yarn.py`. Also, I'll post an RFC to retire Python 2 support.

PR #564, opened 2019-09-24
Make yarn.py compatible with Python 3

hcho3:
Closes dmlc/dmlc-core#538

@ericangelokim @thvasilo Can you review?

### DIFF
--- a/tracker/dmlc_tracker/yarn.py
+++ b/tracker/dmlc_tracker/yarn.py
@@ -5,6 +5,7 @@ dmlc will run as a Yarn application
 # pylint: disable=invalid-name, too-many-locals, too-many-branches, missing-docstring
 from __future__ import absolute_import
 import os
+import sys
 import subprocess
 import warnings
 import logging
@@ -12,6 +13,18 @@ import platform
 from threading import Thread
 from . import opts
 from . import tracker
+
+# Compatibility shim for handling strings
+PY3 = (sys.version_info[0] == 3)
+
+if PY3:
+    def py_str(x):
+        """convert c string back to python string"""
+        return x.decode('utf-8')
+else:
+    def py_str(x):
+        """convert c string back to python string"""
+        return x
 
 def yarn_submit(args, nworker, nserver, pass_env):
     """Submission function for YARN."""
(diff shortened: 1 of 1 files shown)


ericangelokim:
Does this need to be done in any other files?

What would be the repurcussions of just moving this to be only python3 compaible? Py2 is deprecated; it feels backwards to have to support this. Current users can continue to use an older git commit if necessary.

hcho3:
@ericangelokim So you are okay with dropping Python 2.x?

ericangelokim:
The versioning of this code is dependent on xgb version?

hcho3:
@ericangelokim It's other way around. Each XGBoost version has reference to exact commit hash of dmlc-core.

ericangelokim:
Right so existing customers of xgboost (locked to a previous version) will not be affected by this.

In that case I think it makes sense to deprecate python 2 support.

ericangelokim:
You may want to run an example training with python3 when changes are done to ensure that this is the only place that is affected.

thvasilo:
I encountered the same error in the [MPI tracker](https://github.com/dmlc/dmlc-core/blob/master/tracker/dmlc_tracker/mpi.py#L28) today, assume it's the same cause.

hcho3:
@thvasilo I'm not sure why that line would cause issue. `b'mpich' in out` seems quite valid.

thvasilo:
I'll need to check again, I'm running 0.72 and got this error, didn't
notice the new code on master.

thvasilo:
> I encountered the same error in the MPI tracker today, assume it's the same cause.

Yup never mind, this was back in 0.72

thvasilo:
I launched an EMR cluster today to test this out.

I got this error in one of the containers:

```
Traceback (most recent call last):
  File "./launcher.py", line 81, in <module>
    main()
  File "./launcher.py", line 55, in main
    for f in classpath.split(':'):
TypeError: a bytes-like object is required, not 'str'
```

Adding the argument `encoding='utf-8'` to the [call to Popen](https://github.com/dmlc/dmlc-core/blob/285550914d64d04909308af729257255b3f715e5/tracker/dmlc_tracker/launcher.py#L52)  in `launcher.py` fixed the issue.

I think the reason is explained in the [docs for Popen](https://docs.python.org/3.6/library/subprocess.html#popen-constructor):

> If encoding or errors are specified, the file objects stdin, stdout and stderr are opened in text mode with the specified encoding and errors, as described above in Frequently Used Arguments. If universal_newlines is True, they are opened in text mode with default encoding. **Otherwise, they are opened as binary streams.**

hcho3:
@thvasilo Can you try my latest fix?
@trivialfis See #568.

jozsi:
I am getting the following error when trying to run `dmlc-submit`:
```
Traceback (most recent call last):
  File "./launcher.py", line 10, in <module>
    from .util import py_str
ModuleNotFoundError: No module named '__main__.util'; '__main__' is not a package
```

Issue #561, opened 2019-09-08
CI tests are failing for dmlc_exception and std_exception


hcho3:
Failure may be related to concurrency issue, or not. I need to dive deep and make a diagnosis.

PR #566, opened 2019-09-25
[FIX] Build process shouldn’t modify source tree

codingforfun:
It’s best practice to not modify the source tree during build time. 
CMake strongly encourages this. 
This PR changes the cmake build process to create the confuigured header file `build_config.h` in the build tree instead of the source tree.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,11 +26,6 @@ set(DMLC_ENABLED_SANITIZERS "address" "leak" CACHE STRING
   "Semicolon separated list of sanitizer names. E.g 'address;leak'. Supported sanitizers are
   address, leak and thread.")
 
-# include path
-set(INCLUDE_ROOT "${CMAKE_CURRENT_SOURCE_DIR}/include")
-set(INCLUDE_DMLC_DIR "${INCLUDE_ROOT}/dmlc")
-# include_directories("${INCLUDE_ROOT}")
-
 set(dmlccore_LINKER_LIBS "")
 
 # Sanitizer
@@ -115,8 +110,8 @@ else()
   set(DMLC_CMAKE_LITTLE_ENDIAN 1)
 endif()
 
-message(STATUS "${CMAKE_LOCAL}/build_config.h.in -> ${INCLUDE_DMLC_DIR}/build_config.h")
-configure_file("${CMAKE_LOCAL}/build_config.h.in" "${INCLUDE_DMLC_DIR}/build_config.h")
+message(STATUS "${CMAKE_LOCAL}/build_config.h.in -> include/dmlc/build_config.h")
+configure_file("cmake/build_config.h.in" "include/dmlc/build_config.h")
 
 # compiler flags
 if(MSVC)
@@ -205,6 +200,7 @@ add_library(dmlc ${SOURCE})
 target_link_libraries(dmlc ${dmlccore_LINKER_LIBS})
 target_include_directories(dmlc PUBLIC
   $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+  $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/include>
   $<INSTALL_INTERFACE:${CMAKE_INSTALL_PREFIX}/include>)
 target_compile_definitions(dmlc PRIVATE -D_XOPEN_SOURCE=700
   -D_POSIX_SOURCE -D_POSIX_C_SOURCE=200809L -D_DARWIN_C_SOURCE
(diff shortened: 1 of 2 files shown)


codingforfun:
Travis fails on some utests build using Makefiles separate from the ones build by CMake. 
Is that really used anymore. Test build with CMake work. I’m not very keen to dive into some ancient stuff if it’s not needed any more.

hcho3:
@codingforfun See dmlc/dmlc-core#561. This is an issue unrelated to your work here, so don't worry about it.

CMakeLists.txt
```diff
@@ -115,8 +110,8 @@ else()
   set(DMLC_CMAKE_LITTLE_ENDIAN 1)
 endif()
 
-message(STATUS "${CMAKE_LOCAL}/build_config.h.in -> ${INCLUDE_DMLC_DIR}/build_config.h")
-configure_file("${CMAKE_LOCAL}/build_config.h.in" "${INCLUDE_DMLC_DIR}/build_config.h")
+message(STATUS "${CMAKE_LOCAL}/build_config.h.in -> include/dmlc/build_config.h")
+configure_file("cmake/build_config.h.in" "include/dmlc/build_config.h")
```
trivialfis: Please use more specific path like `${CMAKE_CURRENT_LIST_DIR}/cmake/` or `${dmlc_SOURCE_DIR}/cmake/`

codingforfun: Why do you think this is necessary? Documentation is very clear about there configure_file expects it’s inputs and outputs when using relative paths. I think this is nothing we need to expect to change in cmake. So basically trying to pin cmake behaviour using a lot of custom variables is somthing that should be avoided in favor of maintainable and readable code.

hcho3: @codingforfun I just checked the documentation for `configure_file()` and it says that all relative paths are interpreted with respect to the value of [`CMAKE_CURRENT_SOURCE_DIR`](https://cmake.org/cmake/help/v3.2/variable/CMAKE_CURRENT_SOURCE_DIR.html#variable:CMAKE_CURRENT_SOURCE_DIR). I'm not sure `CMAKE_CURRENT_SOURCE_DIR` is necessarily equivalent to [`PROJECT_SOURCE_DIR`](https://cmake.org/cmake/help/v3.2/variable/PROJECT_SOURCE_DIR.html).

Here are definitions:

* `CMAKE_CURRENT_SOURCE_DIR`: The path to the source directory currently being processed. This is the full path to the source directory that is currently being processed by cmake.
* `PROJECT_SOURCE_DIR`: Top level source directory for the current project. This is the source directory of the most recent `project()` command.

Currently, the `CMAKE_LOCAL` variable is defined in terms of `PROJECT_SOURCE_DIR`. I think that `PROJECT_SOURCE_DIR` is correct choice here, since we want the source directory of the most recent `project()` command (dmlc-core is included by xgboost via `add_subdirectory()` command.)

codingforfun: `CMAKE_CURRENT_SOURCE_DIR` is always the directory the cmake file lives in from where you call it. So in this case of a CMakeLists living in the top level directory of dmlc, this is the same as `PROJECT_SOURCE_DIR`. Documentation is somewhat unclear here.

codingforfun: The stackoverflow discussion was regarding `CMAKE_SOURCE_DIR` which is a differnent one.

hcho3: @codingforfun Let me try your change on top of XGBoost and see if it compiles.

hcho3: Indeed, it seems that `CMAKE_CURRENT_SOURCE_DIR` is practically same as `PROJECT_SOURCE_DIR` for our use case.

rongou:
This seems to break the xgboost build:
```bash
git clone --recurse-submodules https://github.com/dmlc/xgboost.git
cd xgboost
mkdir build
cd build
cmake ..
make
```
I get the following error:
```bash
In file included from /home/rou/src/xgboost/include/../../dmlc-core/include/dmlc/./logging.h:16,
                 from /home/rou/src/xgboost/include/../../dmlc-core/include/dmlc/io.h:15,
                 from /home/rou/src/xgboost/rabit/include/rabit/./internal/../serializable.h:14,
                 from /home/rou/src/xgboost/rabit/include/rabit/./internal/engine.h:10,
                 from /home/rou/src/xgboost/rabit/include/rabit/rabit.h:48,
                 from /home/rou/src/xgboost/src/common/timer.cc:4:
/home/rou/src/xgboost/include/../../dmlc-core/include/dmlc/././base.h:282:10: fatal error: dmlc/build_config.h: No such file or directory
 #include <dmlc/build_config.h>
          ^~~~~~~~~~~~~~~~~~~~~
compilation terminated.
src/CMakeFiles/objxgboost.dir/build.make:153: recipe for target 'src/CMakeFiles/objxgboost.dir/common/timer.cc.o' failed
make[2]: *** [src/CMakeFiles/objxgboost.dir/common/timer.cc.o] Error 1
CMakeFiles/Makefile2:295: recipe for target 'src/CMakeFiles/objxgboost.dir/all' failed
make[1]: *** [src/CMakeFiles/objxgboost.dir/all] Error 2
Makefile:129: recipe for target 'all' failed
make: *** [all] Error 2
```
Any ideas on what's going on?

hcho3:
@rongou I just tried `git clone --recurse-submodules https://github.com/dmlc/xgboost.git` and I cannot reproduce the problem.

rongou:
Hmm never mind. Somehow it's picking up a separate `dmlc-core` repo on my drive. Once I deleted it the problem went away.

PR #574, opened 2019-10-18
Bugs in launcher.py

wendycwong:
discovered by Michal M

### DIFF
--- a/tracker/dmlc_tracker/launcher.py
+++ b/tracker/dmlc_tracker/launcher.py
@@ -52,7 +52,7 @@ def main():
         (classpath, _) = subprocess.Popen('%s/bin/hadoop classpath' % hadoop_home,
                                           stdout=subprocess.PIPE, shell=True,
                                           env=os.environ).communicate()
-        for f in classpath.split(':'):
+        for f in classpath.decode('utf-8').split(':'):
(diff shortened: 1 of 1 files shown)


tracker/dmlc_tracker/launcher.py
```diff
@@ -52,7 +52,7 @@ def main():
         (classpath, _) = subprocess.Popen('%s/bin/hadoop classpath' % hadoop_home,
                                           stdout=subprocess.PIPE, shell=True,
                                           env=os.environ).communicate()
-        for f in classpath.split(':'):
+        for f in classpath.decode('utf-8').split(':'):
```
szha: assuming the change is for python2, could you verify it for python3 too?

wendycwong: Nope, this is for python3.  If the changes was not made, the wrong path will be included.  I was trying to play with xgboost0.82 off hadoop/yarn.  I was having a hard time getting it to run.  Michal M discovered this bug and hence the proposed fix.

hcho3:
@wendycwong Can you review #564? It addresses the same bug and supports both Python 2 and 3.

hcho3:
Closing in favor of #564

PR #582, opened 2019-12-09
Fix include/dmlc/omp.h for clang

leezu:
`__GOMP_NOTHROW` does not make sense when building with clang. I'm not sure why the previous condition assumes `defined(__ANDROID__) && defined (__clang__)`. It doesn't matter if were building for Android or just building with clang.

CC @larroy

### DIFF
--- a/include/dmlc/omp.h
+++ b/include/dmlc/omp.h
@@ -11,7 +11,7 @@
 #include <omp.h>
 #else
 
-#if defined(__ANDROID__) && defined (__clang__)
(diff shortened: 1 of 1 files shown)


larroy:
Android started using clang after some release of the NDK, not before. I think that's the origin of that. This might break build with older NDKs.

leezu:
So the condition should just be clang? I'm not sure why the android condition was added in the first place then?

larroy:
Because it failed in android. Maybe it should be just clang, don't know by heart now.

PR #584, opened 2019-12-10
Set stack trace depth with DMLC_LOG_STACK_TRACE_DEPTH

szha:
Set stack trace depth with environment variable DMLC_LOG_STACK_TRACE_DEPTH

### DIFF
--- a/include/dmlc/logging.h
+++ b/include/dmlc/logging.h
@@ -97,6 +97,10 @@ inline std::string StackTrace(
 
 #else  // DMLC_LOG_STACK_TRACE is off
 
+inline size_t LogStackTraceLevel() {
+  return 0;
+}
+
 inline std::string demangle(char const* msg_str) {
   return std::string();
 }
--- a/include/dmlc/logging.h
+++ b/include/dmlc/logging.h
@@ -35,6 +35,17 @@ struct Error : public std::runtime_error {
 };
 
 #if DMLC_LOG_STACK_TRACE
+// get stack trace logging depth from env variable.
+inline size_t LogStackTraceLevel() {
+  size_t level;
+  if (auto var = std::getenv("DMLC_LOG_STACK_TRACE_DEPTH")) {
+    if (1 == sscanf(var, "%zu", &level)) {
+      return level + 1;
+    }
+  }
+  return 0;
+}
+
 inline std::string Demangle(char const *msg_str) {
   using std::string;
   string msg(msg_str);
@@ -69,7 +80,9 @@ inline std::string StackTrace(
   std::ostringstream stacktrace_os;
   std::vector<void*> stack(stack_size);
   int nframes = backtrace(stack.data(), static_cast<int>(stack_size));
-  stacktrace_os << "Stack trace:\n";
+  if (start_frame < static_cast<size_t>(nframes)) {
+    stacktrace_os << "Stack trace:\n";
+  }
   char **msgs = backtrace_symbols(stack.data(), nframes);
   if (msgs != nullptr) {
     for (int frameno = start_frame; frameno < nframes; ++frameno) {
@@ -396,7 +409,7 @@ class LogMessageFatal : public LogMessage {
  public:
   LogMessageFatal(const char* file, int line) : LogMessage(file, line) {}
   ~LogMessageFatal() {
-    log_stream_ << "\n" << StackTrace() << "\n";
+    log_stream_ << "\n" << StackTrace(1, LogStackTraceLevel()) << "\n";
     abort();
   }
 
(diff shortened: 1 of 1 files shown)


szha:
@dmlc/dmlc-core-committer

szha:
Project owners whose projects depend on the logging of dmlc-core, this PR will change the default behavior of exception logging to not print stack trace. It needs to be explicitly turned on by setting DMLC_LOG_STACK_TRACE_DEPTH to be larger than 0. (previous default behavior is equivalent to `DMLC_LOG_STACK_TRACE_DEPTH=10`

comaniac:
I think it's fine tho. It means after this commit the logging will not print the stack trace by default, which makes the messages clearer to end users who don't really know what to do with the stack trace (more importantly, they usually don't touch the build config unless a tutorial tells them to do so). Meanwhile, developers can still turn it on when necessary. Thus, this behavior makes more sense in my opinion.

junrushao1994:
@comaniac I agree with Cody that it is totally fine in this sense: Developers do not expect themselves to see full stack trace. However, due to the fact that it is a breaking change, I think that it is somehow important to leave **compile time** flags for downstream projects to keep consistent behavior.

junrushao1994:
To be clear, I think I am in favor of this idea, because end users won't care about how the c++ library is implemented - this only causes confusion, but no gain of information.

The only reason that I am a little bit concerned is that it is **breaking** change in a centralized upstream project, on which many downstream projects may rely. For now, I am not 100% sure that all downstream projects are willing to do breaking changes at the moment. Therefore, I would say leaving a **compile-time** flag for those projects to avoid breaking behavior would be super helpful - and not hard, right?

include/dmlc/logging.h
```diff
@@ -35,6 +35,17 @@ struct Error : public std::runtime_error {
...
 #if DMLC_LOG_STACK_TRACE
+// get stack trace logging depth from env variable.
+inline size_t LogStackTraceLevel() {
+  size_t level;
+  if (auto var = std::getenv("DMLC_LOG_STACK_TRACE_DEPTH")) {
+    if (1 == sscanf(var, "%zu", &level)) {
+      return level + 1;
+    }
+  }
+  return 0;
```
hcho3:
```suggestion
  return DMLC_LOG_STACK_TRACE_SIZE;
```

@szha I think this change will preserve existing behavior

hcho3:
@szha Thanks for pointing out the full implication of this pull request. I'd actually advocate for keeping the default stack trace size (DMLC_LOG_STACK_TRACE_SIZE, which is 10) when DMLC_LOG_STACK_TRACE_DEPTH environment variable is not set. The reason is that often end users post the error message they encounter, and a full stacktrace is quite helpful in diagnosis.

comaniac:
I got your concern, and here is a short summary for a comparison. I personally do not have preference so I'll let others comment.

**Environment Variable Approach (current)**: Do not touch the compile process but accept an environment variable to control stack depth.

Pros: Dynamic depth adjustion; More flexible.
Cons: Change the default behavior and it is not straightforward for dependent projects to keep the original behavior (because setting OS environment vairable is not a part of compilation process).

**Compile-Time Approach**: Put the depth variable to the compile config.

Pros: Dependent projects can easily keep the original behavior as their preference.
Cons: Changing the stack depth requries re-compilation.

hcho3:
@comaniac Take a look at my suggested revision. I think it mixes the pros of the two approaches you described.

comaniac:
> @comaniac Take a look at my suggested revision. I think it mixes the pros of the two approaches you described.

LGTM. Thanks. One miner point: Does the [off implementation](https://github.com/dmlc/dmlc-core/pull/584/files/a9de55cb7a32a13acb51748fbffad656f7f367c6#diff-bd684563e7398e78d9d88b34e2444548R101) have to be changed as well?

hcho3:
@comaniac No, when DMLC_LOG_STACK_TRACE (compile flag) is set to off, then no stack trace should be printed. This behavior is consistent to the status quo.

junrushao1994:
Thanks @szha, @comaniac and @hcho3 for the valuable discussion! I think concern from everyone of you guys is valid. And I believe @hcho3's suggesting change helps resolve concerns from all of us :-)

szha:
OK thanks. I committed the suggested change. After this change I can change the behavior on just the MXNet side.

szha:
@hcho3 @junrushao1994 does the latest version look good to you?

junrushao1994:
I think it would satisfy all potential use cases :-) Thanks, Sheng!

include/dmlc/logging.h
```diff
@@ -35,6 +35,17 @@ struct Error : public std::runtime_error {
 };
 
 #if DMLC_LOG_STACK_TRACE
+// get stack trace logging depth from env variable.
+inline size_t LogStackTraceLevel() {
+  size_t level;
+  if (auto var = std::getenv("DMLC_LOG_STACK_TRACE_DEPTH")) {
+    if (1 == sscanf(var, "%zu", &level)) {
```
larroy: you should not use sscanf but instead use stoull http://www.cplusplus.com/reference/string/stoull/

hcho3: What would be the benefit, given that this line already performs error checking?

larroy: While in this case seems fine, it's very easy to open the door for buffer overflows with sscanf. In general not good to see it in the code, there are safer alternatives. I would have a linter rule to prevent it's usage.

PR #586, opened 2019-12-22
Add OpenMP as CMake target

hcho3:
Simplify logic for importing OpenMP into the CMake build by treating OpenMP as a target. This way, we no longer have to set compiler flags to enable OpenMP. See https://cliutils.gitlab.io/modern-cmake/chapters/packages/OpenMP.html for more information.

@dmlc/dmlc-core-committer @trivialfis Please help review. This will be especially useful for Mac OSX users. The old way requires them to use Homebrew GCC, which is a heavy dependency (*). The new way allows them to use Apple Clang (default system compiler) instead. 

(*) See discussion at https://github.com/Homebrew/homebrew-core/pull/43246, where Homebrew maintainers asked to remove GCC dependency from XGBoost.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,6 +43,7 @@ addons:
       - g++-4.8
   homebrew:
     packages:
+      - gcc@8
       - cmake
       - libomp
       - python3
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,8 @@ addons:
       - g++-4.8
   homebrew:
     packages:
-      - gcc@7
+      - cmake
+      - libomp
       - python3
     update: true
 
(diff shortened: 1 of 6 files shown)


hcho3:
Sorry for the noise. I'm now done ironing out all loose ends. This PR is now ready for review.

hcho3:
@trivialfis Do you mean that we should remove unittest_gtest for OSX?

trivialfis:
@hcho3 No.  I see package `gcc@7` is removed, so we are no-longer testing dmlc-core compiled with gcc on OSX.  Just curious whether the test is required for dmlc-core.

hcho3:
@trivialfis I put back gcc. Now dmlc-core is tested against both gcc and Apple Clang.

hcho3:
@larroy

larroy:
sorry I missed this LGTM

PR #605, opened 2020-03-31
[Breaking] Don't push zero when value is missing in CSV parser.

trivialfis:
Currently when the value is missing, CSV parser pushes a default value into row
block, which is zero for all data types.  This defeats the purpose of CSR like
row block and generates false information about the data.  Hence I made a
breaking change instead of an addition parameter.

Fix: dmlc/xgboost#5456 .


trivialfis:
@hcho3

trivialfis:
@hcho3 Could you merge it?  I will open a PR for updating dmlc-core on XGBoost.

PR #610, opened 2020-04-14
Define `omp_in_parallel` stub.

trivialfis:
Related issue: https://github.com/dmlc/xgboost/issues/5518

PR #611, opened 2020-04-22
[Breaking] Remove optional `changed` parameter.

trivialfis:
I made the optional parameter for users to detect whether some parameters have been changed.  So far it doesn't seem to be useful in practice, users can just save the old value and compare it themselves.  Supporting this parameter requires a `Same` function to compare raw memory, which could be uninitialized.  This results in many complains and false alarms from analyzers like clang-tidy, ubsan, valgrind.

### DIFF
--- a/include/dmlc/parameter.h
+++ b/include/dmlc/parameter.h
@@ -171,20 +171,17 @@ struct Parameter {
    * \tparam Container container type
    *
    * \param kwargs map of keyword arguments, or vector of pairs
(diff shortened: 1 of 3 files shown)


trivialfis:
The exact same error on #610 ...

trivialfis:
@hcho3

hcho3:
Will merge this after all tests pass.

trivialfis:
ping @hcho3

PR #615, opened 2020-05-21
[Logging] Add `DMLC_NO_INLINE` to `LogMessageFatal`

junrushao1994:
A follow-up PR of #613 

Below are the benchmarks we did. We use `g++ -fstack-usage` to dump the stack usage information for the target function.

Settings:
* Compiler: g++ 5.4
* OS: Ubuntu 18.04
* Stack size: ulimit -s 8192
* Target function: `virtual tvm::relay::Expr tvm::relay::CommonSubexprEliminator::VisitExpr_(const tvm::relay::CallNode*)`, which originally crashed [this PR](https://github.com/apache/incubator-tvm/pull/5585) on a very deep network

Results:

|     | This patch                     | Patch #613 | No patch dmlc-core |
|---|-----------------------|-----------------|--------------------|
| CHECK      | 544              | 560             | 4656               |
| throw dmlc::Error  |       | 736             | 1264               |
| throw std::runtime_error  |  | 480             | 1008               |
| No checks (no-op)           |     | 480             | 1008               |

CC: @tqchen @trivialfis


include/dmlc/logging.h
```diff
@@ -428,30 +426,42 @@ class LogMessageFatal : public LogMessage {
...
+                 << ": ";
+    }
+    dmlc::Error Finalize() {
+#if DMLC_LOG_BEFORE_THROW
+      LOG(ERROR) << log_stream.str();
+#endif
+      return dmlc::Error(log_stream.str());
+    }
+    DMLC_NO_INLINE static Entry *ThreadLocal() {
+      static thread_local Entry *result = new Entry();
```
junrushao1994, line 459: The introduction of TLS helps to reduce the size of class `dmlc::LogMessageFatal` to zero, and thus enables some opportunities in compiler optimization

hcho3:
@trivialfis Can you review?

include/dmlc/logging.h
```diff
@@ -428,30 +426,42 @@ class LogMessageFatal : public LogMessage {
...
-  LogMessageFatal(const char* file, int line) {
-    log_stream_ << "[" << pretty_date_.HumanDate() << "] " << file << ":"
-                << line << ": ";
+  LogMessageFatal(const char *file, int line) {
+    Entry::ThreadLocal()->Init(file, line);
   }
-  std::ostringstream &stream() { return log_stream_; }
-  ~LogMessageFatal() DMLC_THROW_EXCEPTION {
+  std::ostringstream &stream() { return Entry::ThreadLocal()->log_stream; }
+  DMLC_NO_INLINE ~LogMessageFatal() DMLC_THROW_EXCEPTION {
```
junrushao1994, line 433: The no-inline trick here instructs the compiler not to merge the complicated logic of throwing in the destructor into the caller. This usually wouldn't hurt performance because it is quite a slow path

include/dmlc/logging.h
```diff
@@ -428,30 +426,42 @@ class LogMessageFatal : public LogMessage {
...
-    // hopefully we can do it here
-    // also log the message before throw
-#if DMLC_LOG_BEFORE_THROW
-    LOG(ERROR) << log_stream_.str();
+    Entry::ThreadLocal()->log_stream << "\n"
+                                     << StackTrace(1, LogStackTraceLevel())
+                                     << "\n";
 #endif
-    throw Error(log_stream_.str());
+    throw Entry::ThreadLocal()->Finalize();
```
junrushao1994, line 439: We moved all the logic except `throw` itself to another class, to make sure the unsafe region is minimized

junrushao1994:
`Logging.basics` is failing and I am trying to understand the reason

hcho3:
@junrushao1994 Somehow this line is not being honored:
https://github.com/dmlc/dmlc-core/blob/77df189fdeb1d4628d3218d1d9322fcd84496067/test/unittest/unittest_logging.cc#L2

tqchen:
Merging this for now. We should also migrate the windows test to the action later.

tqchen:
We can refer to the windows pipeline of TVM

tqchen:
https://github.com/apache/incubator-tvm/blob/master/.github/workflows/main.yml#L56

junrushao1994:
Hmmm i am not so sure why it happens. Is there any symbol overriding that happens in `dmlc_unit_tests`? Or is there any amalgamation logic which makes one include ineffective?

tqchen:
I am not too sure either, we should look into the cmakefile to see what was happening

tqchen:
Here is a guess, it could be possible that we links every test into a single app. then the config in https://github.com/dmlc/dmlc-core/blob/master/test/unittest/unittest_logging.cc conflicts with the fatal throw one. Perhaps we should just unify the two tests if that is the case

tqchen:
@junrushao1994 can you open a followup PR to do that?

junrushao1994:
@tqchen to unify those two tests by changing both of their macro to 1?

tqchen:
Yes, you might also need to update the ASSERT_DEATH a bit by changing that to throw

PR #616, opened 2020-05-21
[Bugfix] Fix logging tests

junrushao1994:
Hope this could pass the unittests :-)

@hcho3 @tqchen


junrushao1994:
Looks like MSVC 2013 doesn't recognize `thread_local`

tqchen:
Let us just disable msvc 2013 in here https://github.com/dmlc/dmlc-core/blob/master/appveyor.yml

hcho3:
Yes, let's drop MSVC 2013, since dmlc-core has been upgraded to C++11 standard. We need MSVC 2015+ to compile C++11.

junrushao1994:
Thank you @hcho3 and @tqchen!

PR #617, opened 2020-05-26
Implement an Intrusive Pointer class.

trivialfis:


hcho3:
It looks like a reference-counting pointer. What kind of benefit does this provide compared to `std::shared_ptr<>`?

trivialfis:
My use case is for better memory footprint, quoting from boost intrusive ptr:

> The memory footprint of intrusive_ptr is the same as the corresponding raw pointer;

trivialfis:
Em ... gcc 4.8 is used here.  I thought dmlc core is on c++ 11 now?

hcho3:
Yes, we should be using C++11 everywhere.

trivialfis:
@tqchen @hcho3 What do you think?

hcho3:
Does this PR depend on #618? We should merge #618 first.

trivialfis:
No, I worked around it.

tqchen:
I would certainly recommend shared_ptr when possible, unless there is a usecase where massive amount of ptrs need to be introduced

trivialfis:
Thanks for sharing.  For most of the use cases shared pointer is the canonical way to go.  But when number of pointers is huge, lower memory footprint also means higher runtime performance.  It speeds up the JSON in Xgboost for about 40 percent.  Also I see that TVM also uses a varient of intrusive ptr, so I thought it can be useful to have one implementation in dmlc core as it serves as a common directory.  Also it's header only so it won't cause anything if you don't use it.

tqchen:
In the particular case of JSON, it would be interesting to ask if it is due to deep nesting, and whether that could be avoided(e.g. use json structure to store indirect pointers as opposed to the nested data structure), anyway, if we want to introduce intrusive ptr, then we should get it properly reviewed

include/dmlc/intrusive_ptr.h
```diff
@@ -0,0 +1,219 @@
...
+    return *this;
+  }
+
+  void Reset() { Decref(ptr_); }
+  void Reset(element_type *that) { IntrusivePtr{that}.swap(*this); }
+  void Reset(element_type *that, bool add_ref) { IntrusivePtr{that, add_ref}.swap(*this); }
+
+  element_type &operator*() const noexcept { return *ptr_; }
+  element_type *operator->() const noexcept { return ptr_; }
+  element_type *Get() const noexcept { return ptr_; }
```
tqchen, line 153: Make API consistent with the shared_ptr, Get->get

include/dmlc/intrusive_ptr.h
```diff
@@ -0,0 +1,219 @@
...
+
+  explicit operator bool() const noexcept { return static_cast<bool>(ptr_); }
+
+  /*
+   * \brief Helper function for swapping 2 pointers.
+   */
+  void swap(IntrusivePtr &that) noexcept {  // NOLINT
+    element_type *t = ptr_;
+    ptr_ = that.ptr_;
```
tqchen, line 162: std::swap(ptr, that.ptr);

include/dmlc/intrusive_ptr.h
```diff
@@ -0,0 +1,219 @@
...
+  IntrusivePtr &operator=(IntrusivePtr &&that) {
+    std::swap(ptr_, that.ptr_);
+    return *this;
+  }
+  template <typename U> IntrusivePtr &operator=(IntrusivePtr<U> &&that) {
+    std::swap(ptr_, that.ptr_);
+    return *this;
+  }
+
+  void Reset() { Decref(ptr_); }
```
tqchen, line 147: need to set ptr to nullptr.

include/dmlc/intrusive_ptr.h
```diff
@@ -0,0 +1,219 @@
...
+  void Reset(element_type *that) { IntrusivePtr{that}.swap(*this); }
+  void Reset(element_type *that, bool add_ref) { IntrusivePtr{that, add_ref}.swap(*this); }
```
tqchen, line 149: avoid non-stl complaint API, add_ref is dangerous

include/dmlc/intrusive_ptr.h
```diff
@@ -0,0 +1,219 @@
...
+    that.ptr_ = nullptr;
+  }
+
+  ~IntrusivePtr() { Decref(ptr_); }
+
+  IntrusivePtr &operator=(element_type *that) {
+    *this = IntrusivePtr(that);
+    return *this;
+  }
+  IntrusivePtr &operator=(IntrusivePtr const &that) {
```
tqchen, line 121: use https://en.wikibooks.org/wiki/More_C%2B%2B_Idioms/Copy-and-swap

include/dmlc/intrusive_ptr.h
```diff
@@ -0,0 +1,219 @@
...
+
+  IntrusivePtr() noexcept = default;
+
+  /*!
+   * \brief Contruct an IntrusivePtr.
+   *
+   * \param p Raw pointer to object
+   * \param add_ref Whether contructing the IntrusivePtr should increase reference count.
+   */
+  explicit IntrusivePtr(element_type *p, bool add_ref = true) : ptr_{p} {
```
tqchen, line 95: Do not introduce a raw ptr API, instead create a make_intrusive<T> which creates the intrusive ptr.

include/dmlc/intrusive_ptr.h
```diff
@@ -0,0 +1,219 @@
...
+
+namespace dmlc {
+/*!
+ * \brief Helper class for embedding reference counting into client objects.
+ */
+class RefCount {
+  std::atomic<int32_t> count_;
+
+ public:
+  RefCount() noexcept : count_{0} {}
```
tqchen, line 21: Make all these operations private and friend to IntrusivePtr

include/dmlc/intrusive_ptr.h
```diff
@@ -0,0 +1,219 @@
...
+
+#include <atomic>
+#include <cinttypes>
+#include <functional>
+
+namespace dmlc {
+/*!
+ * \brief Helper class for embedding reference counting into client objects.
+ */
+class RefCount {
```
tqchen, line 17: IntrusivePtrCell

tqchen:
I left some feedbacks, if we want to introduce it to the common header, let us make sure we do it right, https://github.com/apache/incubator-tvm/blob/master/include/tvm/runtime/object.h#L349 might be a good reference.

In the particular case of tvm, we need to support assive amount of AST and a cross language ABI. That may not apply to the xgboost's case

trivialfis:
Actually now that I think about it, add `make_intrusive` might not be a good idea as destroying the object is left for a user defined function.  @tqchen Can we use raw pointer constructor instead?

tqchen:
@trivialfis can you given an example why a user defined function is needed? while it is certainly possible to store the deleter inside the intrusive ptr for customized deletion, it becomes more like a `shared_ptr`. In the meantime, if there is a need for customized deleter, shall we go with the alternative [arena](https://github.com/apache/incubator-tvm/blob/master/src/support/arena.h#L88) pattern instead?

trivialfis:
User defined destroy function is removed.

hcho3:
CI is fixed in #622

trivialfis:
@tqchen Could you please take another look?

trivialfis:
ping @tqchen

trivialfis:
@tqchen Aside from the transient failure in thread group tests, is there anything else I need to do for this PR?

hcho3:
@junrushao1994 What do you think of having an intrusive pointer in dmlc-core? Would you like to take a look?

junrushao1994:
Hey @hcho3 and @trivialfis, thank you for mentioning me! Sure thing, I will take a look tomorrow.

It is good to have intrusive pointer infra somewhere in dmlc, so that we can customize them as we want.

In TVM runtime, there is an Object/ObjectPtr/ObjectRef system, which handles both runtime type info and intrusive pointer without having to depend on the compiler, capable of runtime reflection and provides POD C compatible memory layout.

In MXNet, I didn’t see such infra around, so it might be helpful. CC @szha.

szha:
@tqchen any concern for merging?

include/dmlc/intrusive_ptr.h
```diff
@@ -0,0 +1,259 @@
...
+
+namespace dmlc {
+/*!
+ * \brief Helper class for embedding reference counting into client objects.
+ */
+class IntrusivePtrCell {
+  std::atomic<int32_t> count_;
+  template <typename T> friend class IntrusivePtr;
+
+  std::int32_t Inc() noexcept { return ++count_; }
```
tqchen, line 21: Inc -> IncRef, consider add memory order requirement as in https://github.com/apache/incubator-tvm/blob/master/include/tvm/runtime/object.h#L772

include/dmlc/intrusive_ptr.h
```diff
@@ -0,0 +1,259 @@
...
+#include <atomic>
+#include <cinttypes>
+#include <functional>
+
+namespace dmlc {
+/*!
+ * \brief Helper class for embedding reference counting into client objects.
+ */
+class IntrusivePtrCell {
+  std::atomic<int32_t> count_;
```
tqchen, line 18: explicitly mark private

include/dmlc/intrusive_ptr.h
```diff
@@ -0,0 +1,259 @@
...
+      }
+    }
+  }
+
+  /*!
+   * \brief Contruct an IntrusivePtr.
+   *
+   * \param p Raw pointer to object
+   */
+  explicit IntrusivePtr(T *p) : ptr_{p} {
```
tqchen, line 73: Disallow constructor from raw ptr as it is unsafe, at least this should be made as a private function, and use something like IntrinsivePtr<T>::FromRaw

include/dmlc/intrusive_ptr.h
```diff
@@ -0,0 +1,259 @@
...
+  IntrusivePtr &operator=(IntrusivePtr &&that) {
+    std::swap(ptr_, that.ptr_);
+    return *this;
+  }
+  template <typename U> IntrusivePtr &operator=(IntrusivePtr<U> &&that) {
+    std::swap(ptr_, that.ptr_);
+    return *this;
+  }
+
+  void Reset() {
```
tqchen, line 140: use reset to be consistent with STL

include/dmlc/intrusive_ptr.h
```diff
@@ -0,0 +1,259 @@
...
+    std::swap(copy, *this);
+    return *this;
+  }
+  template <typename U> IntrusivePtr &operator=(IntrusivePtr<U> const &that) {
+    IntrusivePtr<element_type> copy{that.get()};
+    std::swap(*this, copy);
+    return *this;
+  }
+
+  IntrusivePtr &operator=(IntrusivePtr &&that) {
```
tqchen, line 131: IntrusivePtr<T>

include/dmlc/intrusive_ptr.h
```diff
@@ -0,0 +1,259 @@
...
+    std::swap(ptr_, that.ptr_);
+    return *this;
+  }
+  template <typename U> IntrusivePtr &operator=(IntrusivePtr<U> &&that) {
```
tqchen, line 135: This is unsafe, unless there is a clear relation between U and T.

tqchen:
Some additional comments. if we really want to do it, let us make sure we polish the impl further, and make sure the API itself is safe to use(e.g. no unsafe interactions with raw ptrs, and different types)

trivialfis:
@tqchen No problem.  I would like to get it right.

include/dmlc/intrusive_ptr.h
trivialfis: ?

include/dmlc/intrusive_ptr.h
trivialfis: This is private.

include/dmlc/intrusive_ptr.h
trivialfis: Removed.

trivialfis:
@tqchen Could you please take another look?

trivialfis:
@tqchen Could you please take another look?

trivialfis:
Any update?

include/dmlc/intrusive_ptr.h
```diff
@@ -0,0 +1,241 @@
...
+  }
+  void reset(element_type *that) { IntrusivePtr{that}.swap(*this); }  // NOLINT
+
+  element_type &operator*() const noexcept { return *ptr_; }
+  element_type *operator->() const noexcept { return ptr_; }
+  element_type *get() const noexcept { return ptr_; }  // NOLINT
+
+  explicit operator bool() const noexcept { return static_cast<bool>(ptr_); }
+
+  int32_t UseCount() noexcept {
```
tqchen, line 134: make API consistent with std::shared_ptr

include/dmlc/intrusive_ptr.h
```diff
@@ -0,0 +1,241 @@
...
+  IntrusivePtr(IntrusivePtr &&that) : ptr_{that.ptr_} { that.ptr_ = nullptr; }
+
+  ~IntrusivePtr() { Decref(ptr_); }
+
+  IntrusivePtr<T> &operator=(element_type *that) {
+    *this = IntrusivePtr(that);
+    return *this;
+  }
+  IntrusivePtr<T> &operator=(IntrusivePtr<T> const &that) {
+    auto copy = that;
```
tqchen, line 113: Use copy constructor instead of the implicit assignment, do 

```c++
IntrinsivePtr<T>(that).swap(*this);
```

include/dmlc/intrusive_ptr.h
```diff
@@ -0,0 +1,241 @@
...
+            typename... Args>
+  friend IntrusivePtr<Base> make_intrusive_base(Args &&... args); // NOLINT
+
+  IntrusivePtr() noexcept = default;
+  IntrusivePtr(IntrusivePtr const &that) : ptr_{that.ptr_} { Incref(ptr_); }
+  IntrusivePtr(IntrusivePtr &&that) : ptr_{that.ptr_} { that.ptr_ = nullptr; }
+
+  ~IntrusivePtr() { Decref(ptr_); }
+
+  IntrusivePtr<T> &operator=(element_type *that) {
```
tqchen, line 108: do not expose this function as it is unsafe

include/dmlc/intrusive_ptr.h
```diff
@@ -0,0 +1,241 @@
...
+
+/*
+ * \brief Make an intrusive pointer with Base class.  This function helps to avoid
+ *        defining reference counting function for each derived class.
+ *
+ * \tparam Base The type of base class.
+ * \tparam Sub The type of subclass.
+ * \tparam Args Type for input arguments for constructor.
+ *
+ * \param args Input arguments for constructor.
```
tqchen, line 217: Add an example code base, this function is not intuitive. Can we avoid the use of make_intrusive_base? Perhaps make every intrusive payload class to subclass a base and use https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern

trivialfis: Disabling raw pointer ctor and this function will require all client code to give up on runtime dispatching, which is one of the primary reasons of using a smart pointer.  Do you really want to do that?

trivialfis: If runtime dispatching is not required, users can probably just define a value instead of pointer for most of the use cases.  I think I would prefer having raw pointer ctor for intrusive ptr instead.

tqchen: I do not quite understand your comment, can you provide an example about what is the usecase? If we need to support runtime type downcasting, then that will likely mean we will need to have a type index in the intrusive ptr cell

trivialfis: Runtime polymorphism.  Something like:

``` c++
class Base {
 public:
  IntrusivePtrCell mutable ref;
  friend IntrusivePtrCell &IntrusivePtrRefCount(Base const *t) noexcept {
    return t->ref;
  }

  virtual CallVirtualFunc() = 0;
};
class SubClassA : public Base {
    void CallVirtualFunc() override {
        // do something
    }
};
class SubClassB : public Base {
    void CallVirtualFunc() override {
        // do something
    }
};


intrusive_ptr<Base> ptr_a { new SubClassA };
ptr_a->CallVirtualFunc();

intrusive_ptr<Base> ptr_b { new SubClassB };
ptr_b->CallVirtualFunc();
```

The friend function and ref count are only defined in base class.

trivialfis: @tqchen What do you think?

tqchen: In this case, I think we should do `make_intrusive_ptr<SubclassA>();`, and then allow automatic casting it to `intrusive_ptr<Base>`. It would be great if the API is mostly consistent with std::shared_ptr, or boost::intrusive_ptr

trivialfis: I modeled the implementation after boost, which uses raw ptr for initialization.  I think for intrusive ptr the cleanest way is to just use raw ptr for initialization.  It's well established in c++ smart pointer with ownership transfer.  The `make_xxx_ptr` came later and do have there limitations.

tqchen: I still do not get the limitation of make_intrusive_ptr though, since in this case clearly we could do make_intrusive_ptr<SubclassA> and then assign to intrusive_ptr<Base> and the intention is quite clear without having to introducing a new make_intrusive_base function.

trivialfis: @tqchen The `IntrusivePtrRefCount(T const *t) noexcept` function is only available for base class and not defined for subclass.  It's a friend function.

trivialfis: Using above example with make_intrusive_ptr:

```
/usr/bin/ld: CMakeFiles/dmlc_unit_tests.dir/unittest_intrusive_ptr.cc.o: in function `dmlc::IntrusivePtr<dmlc::SubClassA>::Decref(dmlc::SubClassA*)':
/home/fis/Workspace/XGBoost/xgboost/dmlc-core/include/dmlc/intrusive_ptr.h:67: undefined reference to `dmlc::IntrusivePtrCell& dmlc::IntrusivePtrRefCount<dmlc::SubClassA>(dmlc::SubClassA const*)'
/usr/bin/ld: CMakeFiles/dmlc_unit_tests.dir/unittest_intrusive_ptr.cc.o: in function `dmlc::IntrusivePtr<dmlc::SubClassA>::Incref(dmlc::SubClassA*)':
/home/fis/Workspace/XGBoost/xgboost/dmlc-core/include/dmlc/intrusive_ptr.h:62: undefined reference to `dmlc::IntrusivePtrCell& dmlc::IntrusivePtrRefCount<dmlc::SubClassA>(dmlc::SubClassA const*)'
collect2: error: ld returned 1 exit status
```

tqchen: How about making all the subclasses who want to use the intrusivePtr to subclass the IntrusivePtrCell? this will allow us to avoid a type specific Cell

trivialfis: That doesn't seem to be intuitive for users?  Can we accept raw pointer in constructor and `reset` method like other smart pointers?  I can add a note that intrusive ptr takes ownership. If we can accept raw pointer then `make_intrusive_base` won't be needed.

tqchen: ctor of raw ptr sounds OK, as long as we do not support operator= for raw ptr. 

I do think subclassing is not as bad when comparing to the membership approach, and we might be able to get more access control(e.g. the subclass cannot access the IncRef DecRef function).

trivialfis: I vaguely remember I did something like this before.  Seems to be 2 different ways to achieve the same thing.  I don't have a preference and will stick with current implementation in next push as the code is already there.  But if you have a strong preference of using subclass instead of friend function, I can try it later.

tqchen: The main advantage of subclassing is that we could enable the assignment of intrusive_ptr<SubclassA> to intrusive_ptr<Base> (by casting to the Cell) for normal ptr storage. It also helps us to protect certain fields, e.g. keep the ref counter field private. Please consider refer to the Object class in tvm runtime

trivialfis: @tqchen  Toke a brief look into the Object class in tvm,  I think the boost model of  intrusive ptr is much simpler right?  

I'm not convinced that by inheritance is better.

> It also helps us to protect certain fields, e.g. keep the ref counter field private.

I don't think any public interface class should be using intrusive ptr.

Issue #620, opened 2020-05-30
missing parentheses in macro


Light-of-Hers:
https://github.com/dmlc/dmlc-core/blob/ae98c553992eb9f8039fb6e1d17286c0698985de/include/dmlc/logging.h#L192
It may cause some problems that "x" and "y" are not enclosed in parentheses.
For example, in tvm/src/runtime/ndarray.cc:58 `CHECK_EQ(dtype.bits & (dtype.bits - 1), 0);`

tqchen:
Thanks @Light-of-Hers for reporting the issue, @junrushao1994 can you help to send a patch to fix the issue?

okuvshynov:
I can fix it a little later today

tqchen:
https://github.com/dmlc/dmlc-core/pull/621

tqchen:
Thanks @Light-of-Hers @okuvshynov

PR #621, opened 2020-05-30
[logging] fix CHECK macro.

okuvshynov:
In https://github.com/dmlc/dmlc-core/pull/619 I broke 2 things about CHECK_** (sorry!):
1) missing parentheses, so some expressions would be evaluated incorrectly and might not even compile. (thanks for catching this in https://github.com/dmlc/dmlc-core/issues/620)
2) if expression is provided, it needs to be evaluated just once, otherwise code like CHECK_EQ(x++, 10) will print out wrong message in case of a failure.

To fix that and keep perf win, formatting was reintroduced back to LogCheck##name. It's been moved to separate function though, so, compiler will inline function doing the check, but 'outline' the expensive formatting one, thus, doing call to larger, expensive function only in case of a failure. 

I double-checked that perf win still holds:

Current version:
Loading data: 7.47028 sec
Loading data: 7.72023 sec
Loading data: 7.99132 sec
Loading data: 7.93434 sec
Loading data: 7.06726 sec
Loading data: 7.60153 sec
Loading data: 8.46663 sec
Loading data: 8.31967 sec
Loading data: 7.37396 sec
Loading data: 7.41063 sec

Old, before PR 619:
Loading data: 9.34473 sec
Loading data: 9.47449 sec
Loading data: 9.56089 sec
Loading data: 9.57217 sec
Loading data: 9.22075 sec
Loading data: 8.84541 sec
Loading data: 9.21536 sec
Loading data: 9.67341 sec
Loading data: 8.94046 sec
Loading data: 9.28459 sec

### DIFF
--- a/include/dmlc/logging.h
+++ b/include/dmlc/logging.h
@@ -164,16 +164,22 @@ inline bool DebugLoggingEnabled() {
 
 #ifndef DMLC_GLOG_DEFINED
 
+template <typename X, typename Y>
+std::string* LogCheckFormat(const X& x, const Y& y) {
+  std::ostringstream os;
+  os << " (" << x << " vs. " << y << ") "; /* CHECK_XX(x, y) requires x and y can be serialized to string. Use CHECK(x OP y) otherwise. NOLINT(*) */
+  return new std::string(os.str());
+}
+
 // This function allows us to ignore sign comparison in the right scope.
-// Once dmlc requires c++11, #pragma can be changed to _Pramga, and directives
-// can be moved to CHECK_BINARY_OP definition itself
-#define DEFINE_CHECK_FUNC(name, op)                                \
-  template <typename X, typename Y>                                \
-  DMLC_ALWAYS_INLINE bool LogCheck##name(const X& x, const Y& y) { \
-    return (x op y);                                               \
-  }                                                                \
-  DMLC_ALWAYS_INLINE bool LogCheck##name(int x, int y) {           \
-    return LogCheck##name<int, int>(x, y);                         \
+#define DEFINE_CHECK_FUNC(name, op)                                        \
+  template <typename X, typename Y>                                        \
+  DMLC_ALWAYS_INLINE std::string* LogCheck##name(const X& x, const Y& y) { \
+    if (x op y) return nullptr;                                            \
+    return LogCheckFormat(x, y);                                           \
+  }                                                                        \
+  DMLC_ALWAYS_INLINE std::string* LogCheck##name(int x, int y) {           \
+    return LogCheck##name<int, int>(x, y);                                 \
   }
 
 #pragma GCC diagnostic push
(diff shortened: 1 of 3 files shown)


tqchen:
Please look into the CI error

okuvshynov:
@tqchen  - that's some timer test (2 of them), which didn't do enough iterations:

```
[ RUN      ] ThreadGroup.TimerThread
[5] TIME: 324
test/unittest/unittest_thread_group.cc:205: Failure
Expected: (count) >= (MIN_COUNT_WHILE_SLEEPING), actual: 9 vs 10
[  FAILED  ] ThreadGroup.TimerThread (593 ms)
[ RUN      ] ThreadGroup.TimerThreadSimple
[5] TIME: 306
test/unittest/unittest_thread_group.cc:238: Failure
Expected: (count) >= (MIN_COUNT_WHILE_SLEEPING), actual: 7 vs 10
[  FAILED  ] ThreadGroup.TimerThreadSimple (524 ms)
```

Is that test usually reliable?

okuvshynov:
I cannot reproduce this locally; i'll update this change a few times to add more logging to the test and see what's happening.

okuvshynov:
While i suspect it's not related to my change, something is definitely different on MacOS. I recorded all the times the timer woke up, without printing them out right away to avoid IO costs, and printed all of them at the end after timer ends:

```
[       OK ] ThreadGroup.ThreadLaunchQueueThread (9 ms)
[ RUN      ] ThreadGroup.TimerThread
61 128 163 228 265 354 367 382 405 450 497 
[       OK ] ThreadGroup.TimerThread (541 ms)
[ RUN      ] ThreadGroup.TimerThreadSimple
58 95 160 206 262 311 385 397 407 497 515 
[       OK ] ThreadGroup.TimerThreadSimple (586 ms)
```

So, while it is configured to sleep for 10ms, it seems to be woken up by scheduler much later; sleep_for guarantees that thread will not resume execution for _at_least_ duration, rather than exact duration.

I looked at a few runs of same test on same platform for other PRs, several of them seem to have similar issue, but making it. If we'd sleep for 10 ms exactly, that'd be ~50 iterations, but we see 10-30 here, for example:

https://github.com/dmlc/dmlc-core/pull/616/checks:
```
[       OK ] ThreadGroup.ThreadLaunchQueueThread (12 ms)
[ RUN      ] ThreadGroup.TimerThread
[5] TIME: 240
[10] TIME: 517
[       OK ] ThreadGroup.TimerThread (714 ms)
[ RUN      ] ThreadGroup.TimerThreadSimple
[5] TIME: 204
[10] TIME: 454
```

Given that this effect shows up before any of my changes, i'd suspect it's something with the way MacOS scheduler wakes the thread up after short sleeps.

When i tried to make test run with longer sleeps, it passed ~easily, but was not really doing 50ms sleeps as well:

```
[ RUN      ] ThreadGroup.TimerThread
[5] TIME: 533
[10] TIME: 856
[15] TIME: 1368
[20] TIME: 1998
[25] TIME: 2447
[       OK ] ThreadGroup.TimerThread (2691 ms)
[ RUN      ] ThreadGroup.TimerThreadSimple
[5] TIME: 725
[10] TIME: 1320
[15] TIME: 1873
[20] TIME: 2330
```

okuvshynov:
even 200ms also shows a skew.

```
[5] TIME: 1165
[10] TIME: 2573
[15] TIME: 3908
[       OK ] ThreadGroup.TimerThread (4139 ms)
[ RUN      ] ThreadGroup.TimerThreadSimple
[5] TIME: 1295
[10] TIME: 2562
[15] TIME: 3919
```

tqchen:
Thanks @okuvshynov !

okuvshynov:
@tqchen  - sure, sorry for breaking it.

PR #626, opened 2020-06-10
[logging] fix leak in LOG(FATAL)

okuvshynov:
As trivialfis pointed out, even though this allocation happens right before
fatal is logged, that might be caught, and process might continue
execution.

Use unique_ptr instead.


trivialfis:
@okuvshynov Thanks for fixing the issue!

Issue #552, opened 2019-08-01
Add CMake option to use dynamic HDFS lib


hcho3:
See https://github.com/dmlc/dmlc-core/pull/262#issuecomment-513555466

cc @byronyi @superbobry

PR #631, opened 2020-08-24
Add CMake option to use dynamic HDFS lib

VoVAllen:
DGL as a shared library now meets problem on building with HDFS support

Fix https://github.com/dmlc/dmlc-core/issues/552


hcho3:
I think we need to add "security" channel to Ubuntu apt.

hcho3:
CI fixed in #632

PR #634, opened 2020-08-28
Allow compiling with 4.8.5

BarclayII:
CentOS 7 does not have GCC 5.  This PR allows compiling with GCC 4.8.5+ which also supports C++11.

CentOS 7 is the platform where DGL as well as other `manylinux2014` Python packages are built on.

### DIFF
--- a/include/dmlc/base.h
+++ b/include/dmlc/base.h
@@ -116,10 +116,12 @@
 #define DMLC_USE_FOPEN64 1
 #endif
 
-/// check if g++ is before 5.0
+/// check if g++ is before 4.8.5 (the compiler from CentOS 7)
 #if DMLC_USE_CXX11 && defined(__GNUC__) && !defined(__clang_version__)
-#if __GNUC__ < 5
-#pragma message("Will need g++-5.0 or higher to compile all"           \
+#if __GNUC__ < 4 || \
+    (__GNUC__ == 4 && __GNUC_MINOR__ < 8) || \
+    (__GNUC__ == 4 && __GNUC_MINOR__ == 8 && __GNUC_PATCHLEVEL__ < 5)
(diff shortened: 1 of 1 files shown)


hcho3:
GCC 4.8 does not fully support C++11. For example, [`<regex>` is broken in GCC 4.8.](https://www.reddit.com/r/gcc/comments/3f9aij/til_that_if_you_use_stdregex_in_gcc_48_your_code/).

Have you look into using devtoolset? I regularly use CentOS 6 to generate `manylinux2010` packages with GCC 5.
Example of devtoolset-4:
* https://github.com/dmlc/xgboost/blob/b3d2e7644ab983787203271d7aa4cb19992288c1/tests/ci_build/Dockerfile.gpu_build_centos6
* https://github.com/dmlc/xgboost/blob/92913aaf7f9e0ee7611c808edf4a57e8a48356ce/tests/ci_build/Dockerfile.jvm

hcho3:
Actually, it might be okay as long as we add a check to ensure that `<regex>` is not used anywhere in the codebase.

BarclayII:
devtoolset-4 seems to be removed from CentOS 7 SCLo and SCLo-rh repositories.  The oldest version provided is devtoolset-7 (GCC 7), which is incompatible with CUDA 9.0.

hcho3:
@BarclayII You can obtain devtoolset-4 from the vault repository. Here are the examples:
* https://github.com/hcho3/centos7-with-devtoolset-4/blob/mainline/Dockerfile
* https://github.com/hcho3/centos7-with-devtoolset-4/blob/mainline/Dockerfile.gpu

BarclayII:
Many thanks.  Directly downloading from 7.6 repo works.

PR #635, opened 2020-09-02
Batch write for HDFS support

VoVAllen:
The `hdfsWrite` function can only write less than max of `tSize(int32_t)` bytes one-time. (Actually less than max of int32_t due to jvm limits). But the argument of `Write` is `size_t`, which is larger than `tSize`. Therefore add batch write support, following the `Read` function.

Related issue: https://github.com/dmlc/dgl/issues/2106#issuecomment-684804134
cc @ChenAris

### DIFF
--- a/src/io/hdfs_filesys.cc
+++ b/src/io/hdfs_filesys.cc
@@ -57,7 +57,7 @@ class HDFSStream : public SeekStream {
     // upper bound overflow.
     // More about max array size:
     // https://stackoverflow.com/questions/31382531/why-i-cant-create-an-array-with-large-size
-    size_t nmax = static_cast<size_t>(std::numeric_limits<tSize>::max()) / 2;
+    const size_t nmax = static_cast<size_t>(std::numeric_limits<tSize>::max()) / 2;
     while (nleft != 0) {
       tSize ret = hdfsWrite(fs_, fp_, buf, std::min(nleft, nmax));
       if (ret > 0) {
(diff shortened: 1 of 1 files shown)


VoVAllen:
Anyone can help review this?

src/io/hdfs_filesys.cc
```diff
@@ -50,14 +50,25 @@ class HDFSStream : public SeekStream {
 
   virtual void Write(const void *ptr, size_t size) {
     const char *buf = reinterpret_cast<const char*>(ptr);
-    while (size != 0) {
-      tSize nwrite = hdfsWrite(fs_, fp_, buf, size);
-      if (nwrite == -1) {
+    size_t nleft = size;
```
jermainewang, line 53: This line is redundant. Just use `size` is fine here.

src/io/hdfs_filesys.cc
```diff
@@ -50,14 +50,25 @@ class HDFSStream : public SeekStream {
...
+    // When using builtin-java classes to write, the maximum write size
+    // would be limited by the the max array size, which is uncertain
+    // Here I used half of the max limit of tSize(int32_t) as nmax, to avoid
+    // upper bound overflow.
+    // More about max array size:
+    // https://stackoverflow.com/questions/31382531/why-i-cant-create-an-array-with-large-size
+    size_t nmax = static_cast<size_t>(std::numeric_limits<tSize>::max()) / 2;
```
jermainewang, line 60: const size_t

src/io/hdfs_filesys.cc
VoVAllen: I prefer to keep it here to make it consistent with the `Read` implementation

PR #639, opened 2021-01-08
Allow the user to furnish their own CHECK() and LOG() macros

tkonolige:
This make DMLC compatible with libraries that do not want to use DMLC's builtin logging facilities. If `DMLC_USE_LOGGING_LIBRARY` is defined, it will be used in place of the builtin library. This definition must point to an include path like `<my/logging.h>`.

### DIFF
--- a/include/dmlc/base.h
+++ b/include/dmlc/base.h
@@ -13,8 +13,15 @@
 
 /*!
  * \def DMLC_USE_LOGGING_LIBRARY
- * Whether to use user-defined logging library. Change this to an
- * include path like <my/lib/logging.h> to replace DMLC's builtin logging.
+ * Whether to use user-defined logging library. If defined, dmlc will not define
+ * the macros CHECK() and LOG() and instead locate CHECK() and LOG()
+ * from the value of DMLC_USE_LOGGING_LIBRARY. The
+ * DMLC_USE_LOGGING_LIBRARY macro shall be of form <my_logging.h>:
+ *
+ * #define DMLC_USE_LOGGING_LIBRARY <my_logging.h>
+ *
+ * Make sure to define CHECK() and LOG() macros in the provided header;
+ * otherwise the build will fail.
  */
 
 /*!
--- a/include/dmlc/base.h
+++ b/include/dmlc/base.h
@@ -12,6 +12,12 @@
 #endif
 
 /*!
+ * \def DMLC_USE_LOGGING_LIBRARY
+ * Whether to use user-defined logging library. Change this to an
+ * include path like <my/lib/logging.h> to replace DMLC's builtin logging.
+ */
+
+/*!
  * \brief whether throw dmlc::Error instead of
  *  directly calling abort when FATAL error occured
  *  NOTE: this may still not be perfect.
--- a/include/dmlc/base.h
+++ b/include/dmlc/base.h
@@ -11,12 +11,12 @@
 #define DMLC_USE_GLOG 0
 #endif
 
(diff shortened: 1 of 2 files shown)


include/dmlc/base.h
```diff
@@ -11,6 +11,12 @@
 #define DMLC_USE_GLOG 0
 #endif
 
+/*!
+ * \def DMLC_USE_LOGGING_LIBRARY
```
tqchen, line 15: seems will cause doxygen warning issue, consider update

tkonolige: Not sure the best way to do this.
```
#ifndef DMLC_USE_LOGGING_LIBRARY
#define DMLC_USE_LOGGING_LIBRARY
#undef DMLC_USE_LOGGING_LIBRARY
#endif
```

Does that seem reasonable?

tqchen: How about we leave out the doxygen \def macro for now and make a generic comment here. Alternatively, comment about this option in the \file part of the logging.h

hcho3 (commented):
There is already a flag `DMLC_LOG_CUSTOMIZE` that lets you define a custom logging callback function `dmlc::CustomLogMessage::Log`. Can you explain how this pull request is different?

tkonolige:
@hcho3 `DMLC_LOG_CUSTOMIZE` only allows customization of `LOG(INFO)` and only allows the user to change the method in which the message is printed. My use case is changing the formatting of the log message (specifically the backtrace portion) which is not possible with `DMLC_LOG_CUSTOMIZE`.

hcho3:
@tkonolige Does this PR remove the definition of `LOG` completely? Is the user expected to furnish a new implementation for `LOG` macro from scratch?

hcho3:
Related: https://discuss.tvm.apache.org/t/rfc-line-numbers-in-backtraces/8711. After reading the thread, I was able to better understand the motivation of this PR.

tkonolige:
@hcho3 exactly. The user will have to implement their own LOG and CHECK functionality.

Thanks for linking that thread. It provides good context.

include/dmlc/base.h
```diff
@@ -11,6 +11,12 @@
...
+ * Whether to use user-defined logging library. Change this to an
+ * include path like <my/lib/logging.h> to replace DMLC's builtin logging.
```
hcho3, line 17:
```suggestion
 * Whether to use user-defined logging library. If defined, dmlc will not define
 * the macros CHECK() and LOG() and instead locate CHECK() and LOG()
 * from the value of DMLC_USE_LOGGING_LIBRARY. The
 * DMLC_USE_LOGGING_LIBRARY macro shall be of form <my_logging.h>.
 * Make sure to define CHECK() and LOG() macros in the provided header;
 * otherwise the build will fail.
```

hcho3 (changes requested):
I revised the comment to clarify the exact semantics of `DMLC_USE_LOGGING_LIBRARY`. What do you think?

hcho3:
@trivialfis You might be interested in this PR. It might help simplify the logging system in XGBoost.

tkonolige:
@hcho3 Thanks for the feedback, I've incorporated your changes to the documentation.

tqchen:
@hcho3 @junrushao1994 please take another look, will merge in a day if no further comment

junrushao1994 (approved):
Yeah I think the comments are quite clear now :-)

hcho3 (approved):

PR #641, opened 2021-02-04
Add macros for OpenMP exception handling

ldesreumaux:
See https://github.com/dmlc/xgboost/issues/6664.

This PR adds macros that can be used to collect and rethrow exceptions thrown in OpenMP regions.

### DIFF
--- a/include/dmlc/common.h
+++ b/include/dmlc/common.h
@@ -48,7 +48,7 @@ inline size_t HashCombine<size_t>(size_t key, const size_t& value) {
 }
 
 /*!
- * \brief OMP Exception class catches, saves and rethrows exception from OMP blocks
+ * \brief OMP Exception class captures and rethrows exception from OMP blocks
  */
 class OMPException {
  private:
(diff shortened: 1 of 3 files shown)


include/dmlc/common.h
```diff
@@ -59,22 +59,12 @@ class OMPException {
 
  public:
   /*!
-   * \brief Parallel OMP blocks should be placed within Run to save exception
+   * \brief should be called in a catch clause to capture the exception
    */
-  template <typename Function, typename... Parameters>
-  void Run(Function f, Parameters... params) {
```
hcho3, line 65: Please do not remove the existing facility for exception catching (`Run` and `omp_exc_`). This facility is used in MXNet and TVM, and removing it will necessitate an extensive code refactoring in these projects. (Keep in mind that dmlc-core is not just used by XGBoost but by other projects as well. So backward compatibility is important.) Instead, add `OMP_INIT()`, `OMP_BEGIN()`, and `OMP_END()` as convenience macros that build on the existing facility.

hcho3 (changes requested):

include/dmlc/common.h
hcho3: @junrushao1994 @szha for visibility

PR #645, opened 2021-03-17
Allow GNU Make user to opt out of using execinfo.h via flag DMLC_LOG_STACK_TRACE

apivovarov:
Some platforms (e.g. OpenWRT) do not have `execinfo.h` in their SDK.

We should check that `cxxabi.h` and  `execinfo.h` are present before enabling `DMLC_LOG_STACK_TRACE` by default in [build_config_default.h](https://github.com/dmlc/dmlc-core/blob/main/include/dmlc/build_config_default.h#L25)

This check is already done in [cmake/build_config.h.in](https://github.com/dmlc/dmlc-core/blob/main/cmake/build_config.h.in#L13).
To be consistent we should also do it in [build_config_default.h](https://github.com/dmlc/dmlc-core/blob/main/include/dmlc/build_config_default.h#L25)

### DIFF
--- a/include/dmlc/build_config_default.h
+++ b/include/dmlc/build_config_default.h
@@ -22,15 +22,17 @@
      && !(defined __MINGW64__) && !(defined __ANDROID__))\
      && !defined(__CYGWIN__) && !defined(__EMSCRIPTEN__)\
      && !defined(__RISCV__) && !defined(__hexagon__)
-#if DMLC_CXXABI_H_PRESENT && DMLC_EXECINFO_H_PRESENT
+  #if defined(DMLC_LOG_STACK_TRACE) && DMLC_LOG_STACK_TRACE
+  #define DMLC_EXECINFO_H <execinfo.h>
+  #else
+  #define DMLC_EXECINFO_H
+  #endif
   #ifndef DMLC_LOG_STACK_TRACE
   #define DMLC_LOG_STACK_TRACE 1
   #endif
   #ifndef DMLC_LOG_STACK_TRACE_SIZE
   #define DMLC_LOG_STACK_TRACE_SIZE 10
   #endif
-  #define DMLC_EXECINFO_H <execinfo.h>
-#endif
 #endif
 
 /* default logic for detecting existence of nanosleep() */
--- a/include/dmlc/build_config_default.h
+++ b/include/dmlc/build_config_default.h
@@ -22,6 +22,7 @@
      && !(defined __MINGW64__) && !(defined __ANDROID__))\
      && !defined(__CYGWIN__) && !defined(__EMSCRIPTEN__)\
      && !defined(__RISCV__) && !defined(__hexagon__)
+#if DMLC_CXXABI_H_PRESENT && DMLC_EXECINFO_H_PRESENT
   #ifndef DMLC_LOG_STACK_TRACE
   #define DMLC_LOG_STACK_TRACE 1
   #endif
@@ -29,6 +30,7 @@
   #define DMLC_LOG_STACK_TRACE_SIZE 10
   #endif
   #define DMLC_EXECINFO_H <execinfo.h>
+#endif
 #endif
 
 /* default logic for detecting existence of nanosleep() */
--- a/include/dmlc/build_config_default.h
+++ b/include/dmlc/build_config_default.h
@@ -22,13 +22,15 @@
      && !(defined __MINGW64__) && !(defined __ANDROID__))\
      && !defined(__CYGWIN__) && !defined(__EMSCRIPTEN__)\
      && !defined(__RISCV__) && !defined(__hexagon__)
-  #if defined(DMLC_LOG_STACK_TRACE) && DMLC_LOG_STACK_TRACE
-  #define DMLC_EXECINFO_H <execinfo.h>
+  #if !defined(DMLC_LOG_STACK_TRACE)
+    #define DMLC_LOG_STACK_TRACE 1
+    #define DMLC_EXECINFO_H <execinfo.h>
   #else
-  #define DMLC_EXECINFO_H
-  #endif
-  #ifndef DMLC_LOG_STACK_TRACE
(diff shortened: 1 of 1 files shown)


include/dmlc/build_config_default.h
```diff
@@ -22,6 +22,7 @@
      && !(defined __MINGW64__) && !(defined __ANDROID__))\
      && !defined(__CYGWIN__) && !defined(__EMSCRIPTEN__)\
      && !defined(__RISCV__) && !defined(__hexagon__)
+#if DMLC_CXXABI_H_PRESENT && DMLC_EXECINFO_H_PRESENT
```
hcho3, line 25: Are users expected to define these flags manually in their builds? I prefer to use opt-out when it comes to the use of `execinfo.h`, since most of the common platforms have `execinfo.h`; users of OpenWRT and the like should be asked to set a flag to opt out.

hcho3 (changes requested):

include/dmlc/build_config_default.h
hcho3: Also keep in mind that `build_config_default.h` is only used if the user is using GNU Make for their build. If CMake is used, then `cmake/build_config.h.in` is used and `build_config_default.h` is discarded.

Consequently, if the user is using GNU Make, then we can't expect the flag `DMLC_EXECINFO_H_PRESENT ` to be readily available.

apivovarov: Flags are set automatically by cmake - `CMakeLists.txt` has the following:
```
check_include_file_cxx(cxxabi.h DMLC_CXXABI_H_PRESENT)
...
find_package(Backtrace)
if(Backtrace_FOUND)
  set(DMLC_EXECINFO_H_PRESENT 1)
```

About `DMLC_LOG_STACK_TRACE=0`. The following line in `dmlc/build_config_default.h` will fail even if DMLC_LOG_STACK_TRACE is 0
```
#define DMLC_EXECINFO_H <execinfo.h>
```

hcho3: > Flags are set automatically by cmake

`dmlc/build_config_default.h` is meant to be ignored when CMake is used. Can you post an example where this is not true?

hcho3: According to this snippet, `dmlc/build_config_default.h` is ignored when `DMLC_CORE_USE_CMAKE` is defined, and dmlc's CMakeLists.txt defines `DMLC_CORE_USE_CMAKE`.
https://github.com/dmlc/dmlc-core/blob/21cc7de0dc9fd6acb796e1be6181fa8e6b6c8f41/include/dmlc/base.h#L331-L335

apivovarov: error happening when I'm building tvm runtime

to reproduce the error `/usr/include/execinfo.h` should be removed
```
git clone --recursive https://github.com/apache/tvm
cd tvm
mkdir build
cd build
cmake ..
make runtime
```
The following file compilation will fail
```
src/runtime/c_runtime_api.cc.o
```

hcho3: It appears that TVM does not call `add_subdirectory(3rdparty/dmlc-core)`, like XGBoost does:
https://github.com/dmlc/xgboost/blob/9c8523432acdce35ffd6a6935f336f587680dddc/CMakeLists.txt#L153

This is the reason why TVM is using `dmlc/build_config_default.h`. This PR in current won't fix the issue, since neither `DMLC_CXXABI_H_PRESENT` nor `DMLC_EXECINFO_H_PRESENT` are defined.

Please see my latest commit for a working fix.

hcho3: @apivovarov I just tried building the latest TVM with my proposed fix. It works well. I just had to add
```cmake
add_definitions(-DDMLC_LOG_STACK_TRACE=0)
```
to the TVM's CMakeLists.txt.

apivovarov: Philip, thank you for xgboost example. I looked at `xgboost/src/c_api/c_api_error.cc` compilation command.
I noticed that it uses:
- `-DDMLC_CORE_USE_CMAKE`
- `-I{xgboost_SOURCE_DIR}/build/dmlc-core/include` # The directory containing `build_config.h`

I compared it with TVM `src/runtime/c_runtime_api.cc` compilation command. It does not have the statements above 

What lines in xgboost `CMakeLists.txt' make it to happened?

I already added to TVM `CMakeLists.txt' the following:
- `add_subdirectory(3rdparty/dmlc-core)`
- `set(TVM_RUNTIME_LINKER_LIBS "dmlc")`
- `target_include_directories(tvm_objs PRIVATE ${tvm_SOURCE_DIR}/3rdparty/dmlc-core/include)`
- `target_include_directories(tvm_runtime_objs PRIVATE ${tvm_SOURCE_DIR}/3rdparty/dmlc-core/include)`

hcho3: @apivovarov Take a look at this line: 
```cmake
target_link_libraries(objxgboost PUBLIC dmlc)
```
https://github.com/dmlc/xgboost/blob/e2d8a994134d25be8318f3e6316a144230c41e5e/CMakeLists.txt#L175

This line propagates all compiler flags from `dmlc` target to `objxgboost` target.

apivovarov: Philip, I fixed TVM build as you recommended (made it similar to xgboost). Not TVM uses `dmlc-core` as a cmake project. https://github.com/apache/tvm/pull/7701

apivovarov: Should we close this PR?

hcho3: I'd like to merge this PR to accommodate users who are using GNU Make on platforms without `execinfo.h`.

hcho3 (approved):

apivovarov:
About TVM PR-7701. TVM uses dmlc-core as a header only library. They do not want to add new "dmlc" cmake target to TVM project and build `libdmlc.a` library. Looks like TVM can only use `build_config_default.h`

PR #649, opened 2021-06-01
Added option to reduce output from cpplint.

Lunderberg:
This passes through the --quiet from lint.py to cpplint.  The main effect is removing the "Done processing" messages that are printed for every C++ file.


Lunderberg:
Closed, functionality implemented in #664.

PR #651, opened 2021-06-17
Add Support for Parquet

bridgream:
I've added support for Parquet files to dmlc-core. I did this in order to enable external memory support for Parquet files in XGBoost. I have tested my code under the XGBoost framework by training two models with identical parameters but using CSV and Parquet files. The two models generate identical predictions on the same test data. (unit test code not included in this pull request)

As my implementation depends on Apache Arrow Parquet, users can optionally enable Parquet integration when building from source with -DUSE_PARQUET=ON. By default this is OFF, so this pull request should not affect users that do not need such support. 

I also implement reading Parquet files with multiple row groups in parallel. Users can set "nthreads=xxx" when passing the URI. It should also fit the `dmlc::data::ThreadedParser` class, but I have not implemented nor tested this yet. 

Any advice or feedback is very much appreciated. Thanks in advance!

@PeterPanonGit @trivialfis


bridgream:
Also @hcho3 from #650. I opened this PR because I figured out the question raised by myself in a previous PR.

trivialfis (commented):
Thanks for working on this!  A few things:
- Please write some tests.
- Address the errors from linter

@hcho3 Do you think it should be part of dmlc-core, or part of xgboost plugin?

bridgream:
@trivialfis Thanks for the comments. Could you please detail the linter errors? I don't see any errors nor warnings with my editor. Still, you might see include errors in `parquet_parser.h` because it depends on Arrow, which may be unresolvable if there is no Arrow installed. This header file will not be included by default and will only compile with `USE_PARQUET=ON`. Are there any other errors you have spotted?

trivialfis:
@hcho3 Could you please help approve the CI?

bridgream:
I wonder if we are still processing this PR? Also, is it possible to squash merge? I forgot to set PR to squash, but I hope to discard commit history. Thanks @trivialfis @hcho3

bridgream:
I was asked to reset commit history and use a new contact information. This branch has clean commit history and is updated to main branch. I apologize for having to create a new PR for that. Could anybody who visits move to #653 thank you!

Issue #565, opened 2019-09-24
[RFC] Drop support for Python 2.x


hcho3:
Python 2.x is reaching end-of-life on December 31, 2019. Many projects, including [dmlc/xgboost](https://github.com/dmlc/xgboost) and [apache/incubator-mxnet](https://github.com/apache/incubator-mxnet) have decided to drop support for Python 2. See https://python3statement.org/

trivialfis:
+1

szha:
MXNet will drop python2 support in the next major version (2.0), so it would be great if dmlc-core can support it until then. Alternatively, dmlc-core can start a new major version in a new branch, while keeping the old branch with python2 support, so that fixes can be ported back to older version.

hcho3:
@szha Okay, let's keep this RFC open until MXNet 2.0 is released.

PR #652, opened 2021-06-23
Fix and support python2.x or python3.9 on yarn.

rryqszq4:
Fix no module named '__main__.util' that not found file util on yarn.

Fix should be classpath and not class_path.

Support python2.x and python3.9 such as change threading.isAlive to is_alive(https://github.com/python/cpython/blob/87109f4d85c93a870ee8aa0d2b394547d4636b17/Doc/whatsnew/3.9.rst#id717),  or change subprocess.call to subprocess.Popen because no args 'env' at method call in python2.x .

### DIFF
--- a/tracker/dmlc_tracker/kubernetes.py
+++ b/tracker/dmlc_tracker/kubernetes.py
@@ -12,10 +12,10 @@ import sys
 import uuid
 import logging
 if "kubernetes" in sys.modules:
-	from kubernetes import client, config
+    from kubernetes import client, config
 from . import tracker
 if "yaml" in sys.modules:
-	import yaml
+    import yaml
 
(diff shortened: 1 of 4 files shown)


szha (commented):
I wonder why there's still a need to support py2 since it has reached end of life.

rryqszq4:
> I wonder why there's still a need to support py2 since it has reached end of life.
@szha Because our cluster env was python2, i try to run on yarn but failed, so i fix some problem.
Just found this issues #565 but it's important for me, so i need fix it.

hcho3:
Closing, as the pull request has been stable. Feel free to re-submit if the change is still necessary. Thanks!

PR #653, opened 2021-06-29
Add parquet support

bridgream:
I was asked to reset commit history and use a new contact information. This branch has clean commit history and is updated to main branch. I apologize for having to create a new PR for that. 

Quoted: 

> I've added support for Parquet files to dmlc-core. I did this in order to enable external memory support for Parquet files in XGBoost. I have tested my code under the XGBoost framework by training two models with identical parameters but using CSV and Parquet files. The two models generate identical predictions on the same test data. (unit test code not included in this pull request)
> 
> As my implementation depends on Apache Arrow Parquet, users can optionally enable Parquet integration when building from source with -DUSE_PARQUET=ON. By default this is OFF, so this pull request should not affect users that do not need such support.
> 
> I also implement reading Parquet files with multiple row groups in parallel. Users can set "nthreads=xxx" when passing the URI. It should also fit the dmlc::data::ThreadedParser class, but I have not implemented nor tested this yet.
> 
> Any advice or feedback is very much appreciated. Thanks in advance!

@PeterPanonGit @trivialfis @hcho3


bridgream:
Accidentally closed. I'm so sorry.

bridgream:
Hi, is there any update on this PR? Please let me know if there is still any work I should do. Thanks in advance @trivialfis @hcho3

src/optional/parquet_parser.h
```diff
@@ -0,0 +1,183 @@
...
+  data->resize(next_row_groups);
+
+  for (int tid = 0; tid < next_row_groups; ++tid) {
+    int row_group_id = row_groups_read_ + tid;
+    threads.push_back(std::thread([this, row_group_id, data, tid] {
+      this->omp_exc_.Run([&] {
+        ParseRowGroup(row_group_id, &(*data)[tid]);
+      });
+    }));
+  }
```
hcho3, line 119: Let's use `std::async` instead, so that exceptions are automatically propagated from worker threads to the main threads.

src/optional/parquet_parser.h
```diff
@@ -0,0 +1,183 @@
...
+  int64_t values_read;
+
+  for (int i_row = 0; i_row < num_rows_this_group; i_row++) {
+    IndexType idx = 0;
+    DType label = DType(0.0f);
+    real_t weight = std::numeric_limits<real_t>::quiet_NaN();
+
+    for (int i_col = 0; i_col < num_cols_; i_col++) {
+      all_float_readers[i_col]->ReadBatch(chunk_size, nullptr, nullptr, &v, &values_read);
+      assert(values_read == chunk_size);
```
hcho3, line 159:
```suggestion
      CHECK_EQ(values_read, chunk_size);
```

hcho3 (changes requested):

src/optional/parquet_parser.h
bridgream: Thank you @hcho3 , I've replaced `std::thread` with `std::async`. As the implementation is borrowed from [dmlc-core/src/data/text_parser.h, lines 124-137](https://github.com/dmlc/dmlc-core/blob/main/src/data/text_parser.h#L124-L137), I wonder if that could be updated as well.

hcho3: Let's leave that to a future pull request.

src/data/text_parser.h
```diff
@@ -111,7 +111,7 @@ inline bool TextParserBase<IndexType, DType>::FillData(
     std::vector<RowBlockContainer<IndexType, DType> > *data) {
   InputSplit::Blob chunk;
   if (!source_->NextChunk(&chunk)) return false;
-  const int nthread = omp_get_max_threads();
+  const int nthread = this->nthread_;
```
hcho3, line 114: This ensures that `nthread` parameter is honored by the text parser. Previously, the text parser would always use all CPU cores.

hcho3:
@trivialfis Can you review the changes I made to the CI?

hcho3 (approved):

test/unittest/dmlc_unittest.mk
```diff
@@ -1,13 +0,0 @@
-UTEST_ROOT=test/unittest
-UNITTEST=$(UTEST_ROOT)/dmlc_unittest
-UNITTEST_SRC=$(wildcard $(UTEST_ROOT)/*.cc)
-UNITTEST_OBJ=$(patsubst %.cc,%.o,$(UNITTEST_SRC))
-
-GTEST_LIB=$(GTEST_PATH)/lib/
-GTEST_INC=$(GTEST_PATH)/include/
-
-$(UTEST_ROOT)/%.o : $(UTEST_ROOT)/%.cc libdmlc.a
-	$(CXX) $(CFLAGS) -I$(GTEST_INC) -o $@ -c $<
```
hcho3, line 10: I deleted `dmlc_unittest.mk`, since it would attempt to build `unittest_parquet_parser.cc` using GNU Makefile. It is just too much hassle to use GNU Make to build a C++ program with Arrow and Parquet.

src/data/parquet_parser.h
```diff
@@ -0,0 +1,189 @@
...
+      .describe("Column index that will put into instance weights.");
+  }
+};
+
+template <typename IndexType, typename DType = real_t>
+class ParquetParser : public ParserImpl<IndexType, DType> {
+ public:
+  ParquetParser(const std::string& filename,
+                const std::map<std::string, std::string>& args,
+                int nthread) : row_groups_read_(0), nthread_(nthread) {
```
trivialfis, line 54: Is the `nthread` necessary?  It's passed as `-1`, then `nthread_` is overridden by the one in `param_`.

src/data/parquet_parser.h
```diff
@@ -0,0 +1,189 @@
...
+  int num_rows_;
+  int num_cols_;
+  int num_row_groups_;
+  int row_groups_read_;
+  // whether we have reached end of parquet file
+  bool have_next_;
+  // number of threads; hardcoded 4 for now
+  int nthread_;
+  // OMPException object to catch and rethrow exceptions in omp blocks
+  dmlc::OMPException omp_exc_;
```
trivialfis, line 100: I don't think you need this one if you are using async.

trivialfis (commented):
Overall looks good to me, will make some additional changes later today.

src/data/text_parser.h
bridgream: I didn't modify this file but this looked like a bug previously introduced. In my practice, CPU usage never exceeds 500% on a 64-core CPU. 
Also, do we need to change this too? https://github.com/dmlc/dmlc-core/blob/main/src/data.cc#L59 It looks like nthreads_ has been hardcoded as 2. Otherwise, users may experience performance drop.

src/data/parquet_parser.h
bridgream: I agree, It's not. In fact, `nthread` first being passed as `-1` is unnecessary. Furthermore, I observe that nthreads for other parsers has been hardcoded as 2 here https://github.com/dmlc/dmlc-core/blob/main/src/data.cc#L59 (also L29 and L44). However, as @hcho3 has pointed out, the effective nthread is defined as `omp_get_max_threads()` here https://github.com/dmlc/dmlc-core/blob/main/src/data/text_parser.h#L114. Maybe we can remove `nthread` for all the parsers. 

In my practice, CPU usage never exceeded 500% on a 64-core CPU, which explains why this bug went unnoticed.

src/data/parquet_parser.h
bridgream: `omp_exc_` is used here https://github.com/bridgream/dmlc-core/blob/parquet_support/src/data/parquet_parser.h#L119 but I'm also not sure whether `omp_exc_` is still needed. 

Or maybe the question is: do we intend to deprecate OpenMP tasks in favor of `std::async`?

test/unittest/dmlc_unittest.mk
bridgream: Acknowledged. Thank you!

CMakeLists.txt
```diff
@@ -67,6 +68,17 @@ if (DMLC_USE_SANITIZER)
   enable_sanitizers("${DMLC_ENABLED_SANITIZERS}")
 endif (DMLC_USE_SANITIZER)
 
+if(USE_PARQUET)
+  find_package(Arrow REQUIRED)
+  find_package(Parquet REQUIRED)
+  if(BUILD_SHARED_LIBS)
+    target_link_libraries(dmlc PRIVATE arrow_shared parquet_shared)
+  else()
+    target_link_libraries(dmlc PUBLIC arrow_shared parquet_shared)
```
bridgream, line 77: Do we need to change `arrow_shared` and `parquet_shared` into `arrow_static` and `parquet_static`, as there is a [static correspondence](https://arrow.apache.org/docs/cpp/cmake.html)?

src/data/parquet_parser.h
trivialfis: > nthread first being passed as -1 is unnecessary

Thanks for the confirmation.

>  Maybe we can remove nthread for all the parsers and allow users to specify it in the URI.

For now we can limit the scope of this PR and just get the parser in first.  The nthread is quite problematic from my experience.  Using more threads means spawning more batches and buffers, leading to memory usage issues and reproducibility issues.  Can we remove this `-1`? and `nthread` in constructor?  Sorry for the inconvience.

src/data/parquet_parser.h
trivialfis: I don't have a preference.  The comment is just, if one uses OpenMP or plain thread, then the guard is needed.  If one uses async as in the current form of this PR, then we can remove the guard.  Pick one method and we call it a day. ;-)

CMakeLists.txt
trivialfis: That would be great!

trivialfis:
@hcho3 The CI change looks good, thanks for working on this.

src/data/parquet_parser.h
bridgream: Thanks for your feedback! I've removed the `nthread` argument and changed the default `nthreads` (passed by URI) to 1 [here](https://github.com/bridgream/dmlc-core/blob/parquet_support/src/data/parquet_parser.h#L44). I use `nthreads` in my project and it has been good so far (with predictable memory usage and reproducible results). Hopefully we can expose this feature to users in the future.

src/data/parquet_parser.h
bridgream: I've experimented and it looks the guard is not needed if no exception is thrown. Maybe we can get rid of `dmlc::OMPException` for now? I've committed a version without `omp_exc_` but can revert it easily as it's the last one.

src/data/text_parser.h
hcho3: I fixed this particular bug because it was breaking the unit test for the Parquet parser. Let's solve other bugs in future pull requests.

trivialfis (approved):
Looks good to me.  Sorry for the slow response and thanks for the new feature!

PR #658, opened 2021-08-23
Move constructor is added to support dmlc::optional<T>

mozga-intel:
## Description ##
The class template std::optional manages an optional contained value, i.e. a value that may or may not be present. In this pull request, a few things were implemented, such as extended series of constructors and a function of value. Mainly, this pull request is a realization of changes that have an impact on constructors, as follows, (upgrade dmlc::optional<T> constructors):

1. constructor to construct an object that does not contain a value,
2. constructor to construct an object that contains a nullptr value,
3. constructs an optional object that contains a value, initialized as if direct-initializing:`
3. move constructor: If other contains a value, then the stored value is direct-intialized with it.
     `optional(optional &&other) noexcept() { }`
4. move constructor: constructs the stored value with value with `other`parameter.
    `optional(U &&other) noexcept { }`

Upgrade: dmlc::optional<T> value
1.` T &value() &`
2. `const T &value() const &`
3. `T &&value() &`
4. `const T && value() const &&`

### DIFF
--- a/include/dmlc/optional.h
+++ b/include/dmlc/optional.h
@@ -72,6 +72,8 @@ class optional {
 public:
   /*! \brief constructs an object that does not contain a value. */
   optional() : is_none(true) {}
+   /*! \brief constructs an object that does contain a nullopt value. */
+  optional(nullopt_t) noexcept : is_none(true) {}
   /*! \brief copy constructor, if other contains a value, then stored value is
    * direct-intialized with it. */
   optional(const optional &other) = default;
--- a/include/dmlc/optional.h
+++ b/include/dmlc/optional.h
@@ -37,7 +37,7 @@ template <bool B, typename T = void>
 using enable_if_t = typename std::enable_if<B, T>::type;
 
 /*! \brief disambiguation tags that can be passed to the constructors of
- * std::optional. A tag type to tell constructor to construct its value in-place
+ * dmlc::optional. A tag type to tell constructor to construct its value in-place
  */
 struct in_place_t {
   explicit in_place_t() = default;
@@ -46,7 +46,7 @@ struct in_place_t {
 static constexpr in_place_t in_place{};
 
 /*! \brief  Is not constructible or convertible from any expression of type
-(possibly const) std::optional<U>, i.e., the following 8 type traits are all
+(possibly const) dmlc::optional<U>, i.e., the following 8 type traits are all
 false: Link: https://en.cppreference.com/w/cpp/utility/optional/optional */
 template <typename T> class optional;
 template <typename T, typename U, typename Other>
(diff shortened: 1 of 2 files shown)


doc/Doxyfile
```diff
@@ -1691,7 +1691,7 @@ LATEX_HIDE_INDICES     = NO
 # The default value is: NO.
 # This tag requires that the tag GENERATE_LATEX is set to YES.
 
-LATEX_SOURCE_CODE      = NO
+# LATEX_SOURCE_CODE      = NO
```
mozga-intel, line 1694: This change is shifted into another pull request: Please have a look at #660.

sfraczek (approved):
LGTM. It would be nice to see if there is a performance boost thanks to supporting move here.

mozga-intel:
@hcho3 If you can review it, I'll be grateful. Thanks!

mozga-intel:
Gentle ping @hcho3, @szha

szha (approved):
LGTM. I will leave it open for a couple more days in case there's any concern before I merge.

hcho3 (approved):
No objection from me. Hopefully we can use `std::optional` in the future.

mozga-intel:
@szha, @hcho3 Thanks! There are a few fings to improve: firstly, a constructor that can be declared with a constexpr specifier. Secondly, as @hcho3 mentioned, std::optional<T> might be added to support it ~ however it'll require a few changes in the code structure.

szha:
@mozga-intel sounds good. Do you intend to continue in this PR or as a follow-up?

mozga-intel:
@szha after merge! Because I need to test it before.

areusch:
not quite sure why yet, but it looks like when i apply this patch to TVM, the following command (in tvm repo, at rev c07a46327c86fc541297ebb985cc9c1dcef5a0db, using docker container tlcpack/ci-cpu:v0.84 and config.cmake generated with `tests/scripts/task_config_build_cpu.sh`) segfaults:
```
cd rust/tvm/examples/resnet
python3 src/build_resnet.py --build-dir=.
```

still tracing through to see why, but posting up here in case any of you all have ideas.

cc @tqchen

include/dmlc/optional.h
```diff
@@ -31,40 +31,148 @@ struct nullopt_t {
...
-  explicit optional(const T& value) {
-#pragma GCC diagnostic push
-#if __GNUC__ >= 6
-#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
-#endif
+  /*! \brief constructs an object that does contain a nullopt value. */
+  optional(dmlc::nullopt_t) noexcept : is_none(true) {} // NOLINT(*)
+  /*! \brief copy constructor, if other contains a value, then stored value is
+   * direct-intialized with it. */
+  optional(const optional &other) = default;
```
areusch, line 85: it seems that reverting the use of the default copy constructor here stops the segfault

PR #665, opened 2022-04-12
fix tracker for python 3.9

rongou:
Python 3.9 removed the deprecated method `Thread.isAlive` and replace it with `Thread.is_alive`.

See https://bugs.python.org/issue37804


rongou:
@hcho3 @trivialfis

trivialfis:
I think XGBoost is no longer using the tracker in dmlc-core.

rongou:
It's still used in the distributed tests: https://github.com/dmlc/xgboost/blob/master/tests/distributed/runtests.sh#L7

trivialfis:
We can copy the submit script into XGBoost as well, otherwise, there are two trackers in use.  The fix looks good to me, @hcho3 can help merge it

PR #666, opened 2022-04-28
Revert copy-constructor modification which causes TVM segfaults

areusch:
See https://github.com/dmlc/dmlc-core/pull/658

Unfortunately I don't have any more information on why this happens, but tvm is now blocked in updating dmlc-core. I had a unittest that could pass previously, but didn't pass after #658--but, a) I unfortunately lost it in auto-updating a submodule b) don't know if that was really the problem because most of these single-arg constructors aren't marked explicit.

cc @tqchen @mozga-intel @szha

### DIFF
--- a/include/dmlc/optional.h
+++ b/include/dmlc/optional.h
@@ -80,9 +80,18 @@ class optional {
   optional() : is_none(true) {}
   /*! \brief constructs an object that does contain a nullopt value. */
   optional(dmlc::nullopt_t) noexcept : is_none(true) {} // NOLINT(*)
-  /*! \brief copy constructor, if other contains a value, then stored value is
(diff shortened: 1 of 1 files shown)


hcho3:
It might be better to consider upgrading to C++17 ?

areusch:
we could do that, but i think we prefer to be able compile `libtvm_runtime.so` with e.g. c++11 in case it is built on targets with a less modern compiler. i'm not sure exactly what this would mean here

areusch:
:/ it seems the dmlc-core CI is busted

tqchen (approved):

tqchen:
Merged since the hotfix is important to make sure optional function correctly(for non-POD types)

PR #672, opened 2022-08-08
Add Necessary Checks and Definitions to Build on QNX

Johnson9009:
Recently we do the TVM deploy work on QNX, and found 2 places that dmlc-core haven't handle correctly.
1. Backtraces
Even through QNX have a library named "libbacktrace" can do the job, but the [QNX official document](http://www.qnx.com/developers/docs/7.0.0/index.html#com.qnx.doc.neutrino.technotes/topic/backtrace.html) notice that it is an unsupported feature, in addition the API is very different with those of file "execinfo.h", so we think maybe just don't do the bacetrace on QNX is a better solution.
3. Endian
Refer to https://github.com/protocolbuffers/protobuf/pull/10080, endian can be detected on QNX correctly.

### DIFF
--- a/include/dmlc/build_config_default.h
+++ b/include/dmlc/build_config_default.h
@@ -21,7 +21,8 @@
      && !defined(__sun) && !defined(__SVR4)\
      && !(defined __MINGW64__) && !(defined __ANDROID__))\
      && !defined(__CYGWIN__) && !defined(__EMSCRIPTEN__)\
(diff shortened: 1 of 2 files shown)


Johnson9009:
@tqchen @mli @hcho3 Please help to approve running workflows, thanks.

Johnson9009:
Is there anybody can help?

PR #673, opened 2022-09-04
Consistently use lowercase names for Windows libraries and headers

giordano:
MinGW consistently uses lowercase names for all libraries and header files, and
this toolchain is often employed to build libraries for Windows on Linux
systems, which typically use case-sensitive file systems.


giordano:
Bump

giordano:
Bump?

PR #678, opened 2023-03-01
Get rid of std::iterator which is deprecated in C++17

rongou:
XGBoost recently moved to C++17 (https://github.com/dmlc/xgboost/pull/8853) which now shows a deprecation warning for calling `std::iterator`. See https://www.fluentcpp.com/2018/05/08/std-iterator-deprecated/.


rongou:
@hcho3

hcho3:
Thanks!

PR #679, opened 2023-03-06
Don't populate labels if label column is not specified in csv parser

rongou:
Right now the CSV parser sets labels to 0 if the label column is not specified (or set to -1). This is surprising to the user and leads to cryptic error messages. It's probably better to just leave the labels as empty if not specified.

For vertical federated learning, we may have workers that don't have access to the label, so this would enable them to parse csv shards without erroneously setting labels to 0.


rongou:
@hcho3 @trivialfis

hcho3:
Merging for now. I'll try to make time to fix the CI.

PR #691, opened 2025-04-01
[CI] Set up linters using pre-commit

hcho3:
Also remove outdated tests and other unused files.


hcho3:
I will continue in a follow-up PRs

